Issue #66, opened 2019-11-03
Indy: Bad indy-key-mgmt Docker image link


juraj-durech:
The image won't be on Docker Hub, we have to point to a private repo (configured in network.yaml) and add to it just the image name and version here

_Originally posted by @juraj-durech in https://github.com/hyperledger-labs/blockchain-automation-framework/pull/57/review_comment/create_

juraj-durech:
Closed by #69

PR #69, opened 2019-11-04
Add docker repository url from network.yaml (#66)

VladimirDzamba:
Signed-off-by: VladimirDzamba <vladimir.dzamba@accenture.com>


platforms/hyperledger-indy/configuration/roles/create/helm_component/templates/crypto-generate.tpl
```diff
@@ -19,7 +19,7 @@ spec:
       name: {{ network.name }}
     image:
       name: {{ component_name }}
-      repository: indy-key-mgmt:latest
+      repository: {{ network.docker.url }}/indy-key-mgmt:latest
```
juraj-durech: We shouldn't use latest, I would stick to the BAF release version - 0.3.0.0

Issue #74, opened 2019-11-05
Indy: Change trustees and stewards to a list


juraj-durech:
Change the network.yaml and Vault structure to a list of trustees and stewards.

Example:

```
      services:
        trustees:
        - trustee:
          name: authority-trustee
```

juraj-durech:
Closed by #82

PR #82, opened 2019-11-05
Changed trustees and endorsers to a list

juraj-durech:
Signed-off-by: Juraj Durech <juraj.durech@accenture.com>


juraj-durech:
Closes #74

Issue #18, opened 2019-10-29
Indy: Ansible role to get EIP metadata from AWS


sownak:
Prepare an Ansible role which retrieves the EIP allocation IDs for all steward public IPs found in the network.yaml and stores them in an Ansible fact / variable to be used later. The role should also check that the same public IP is specified for all stewards within one organization.

juraj-durech:
Closed by #143

PR #143, opened 2019-12-03
Create ansible role to get EIP metadata from AWS and update create am…

VladimirDzamba:
…bassador with this EIP (#18)

Signed-off-by: VladimirDzamba <vladimir.dzamba@accenture.com>

### DIFF
--- a/platforms/shared/configuration/roles/setup/ambassador/tasks/main.yaml
+++ b/platforms/shared/configuration/roles/setup/ambassador/tasks/main.yaml
@@ -27,17 +27,25 @@
   register: terminal
   when: network['type'] == 'indy'
 
+- name: Get Elastic IP
+  environment:
+    AWS_ACCESS_KEY_ID: "{{ aws.access_key }}"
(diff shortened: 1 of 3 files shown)


platforms/shared/configuration/roles/setup/ambassador/tasks/main.yaml
```diff
@@ -27,17 +27,25 @@
   register: terminal
   when: network['type'] == 'indy'
 
+- name: Get Elastic IP
+  environment:
+    AWS_ACCESS_KEY_ID: "{{ aws.access_key }}"
+    AWS_SECRET_ACCESS_KEY: "{{ aws.secret_key }}"
+  shell: |
+    aws ec2 describe-addresses --public-ips {{ item.services.stewards[0].publicIp }} --region {{ aws.region }} --output json | jq '.Addresses[0].AllocationId'
```
juraj-durech: What if each organization is in a different K8S cluster and thus should have an own Ambassador with a specific public IP? Will this setup work in that case?

Issue #211, opened 2020-01-07
Join channel job fails if peer has already joined the channel


jagpreetsinghsasan:
**Describe the bug**
Join channel job in hyperledger fabric playbook fails if the peer is already in the channel

**Expected behavior**
The job shall check the peer status and should run join channel only if it hasn't join

PR #212, opened 2020-01-07
Added patch fix for join channel

jagpreetsinghsasan:
#211


sownak:
Merged for Fix of issue #211

Issue #210, opened 2020-01-06
Create a docker file to build the BAF environment


suvajit-sarkar:
### Description
As a developer I want to create a docker file so that I can build the BAF environment and deploy the BAF network 

Initial work done by Pull request #108 

### Acceptance Criteria
1. Docker runtime independent from target Kubernetes cluster
2. Platform gets deployed

sownak:
Closed as per PR #244

PR #244, opened 2020-01-16
BAF docker build #210

suvajit-sarkar:
#210


suvajit-sarkar:
> I think this needs a readthedocs readme under developer guide (similar to mac one). It is not very clear like where the network.yaml should be stored.

Yes I plan to add that

jagpreetsinghsasan:
The repo cloning/pushing to the container is not mentioned or not clear.

sownak:
> The repo cloning/pushing to the container is not mentioned or not clear.

This should be now clear in the developer guide readme

jagpreetsinghsasan:
It's fine now

Issue #184, opened 2019-12-27
Quorum: Ansible role to configure Hashicorp Vault access


sownak:
### Description
As a developer I want to configure Kubernetes to access Vault so that I can deploy Quorum network on Kubernetes which reads secrets from Vault.
### Sample steps to be automated using Ansible 
1. Namespace, reviewer, clusterrole-binding etc should already be created.
2. `vault auth enable --path="quorumorg2" kubernetes`
3. get REVIEWER_TOKEN eg: 
```
export REVIEWER_TOKEN=$(kubectl get secret $(kubectl get serviceaccount -n org2-ns vault-reviewer -o jsonpath={.secrets[0].name}) -n org2-ns -o  jsonpath={.data.token} | base64 --decode)
```
4. get ca-cert of kubernetes eg: 
```
kubectl get secrets -n org2-ns $(kubectl get secrets -n org2-ns | grep "default" | awk '{print $1}') -o jsonpath="{.data['ca\.crt']}" | base64 --decode > ca.cert
```
5. Get kubernetes host eg: 
```
export private_dns=$(kubectl config view -o jsonpath="{.clusters[?(@.name=='<cluster name context>')].cluster.server}")
```
6. Configure Vault auth eg: 
```
vault write auth/quorumorg2/config token_reviewer_jwt="$REVIEWER_TOKEN" kubernetes_host="$private_dns" kubernetes_ca_cert=@ca.cert
```
7. Create Policy to control access using templates eg: 
```
cat <<EOF > vault-crypto-ro.hcl
path "secret/<namespace>/crypto/*" {
    capabilities = ["read"]
}
EOF
vault write sys/policy/vault-crypto-org2-ro policy="@vault-crypto-ro.hcl"
```
8: Create Role associated with policy in the kubeauth authentication backend
```
vault write auth/quorumorg2/role/vault-role bound_service_account_names=vault-auth bound_service_account_namespaces=org2-ns policies=vault-crypto-org2-ro ttl=120s
```
### Acceptance Criteria
1. Kubernetes path enabled for each org separately.
2. All commands should be idempotent, i.e. check Vault first if paths, access etc. already created, if yes, do not recreate.
3. Follow similar approach to existing code for Corda/Fabric.

sownak:
Closed as per latest pull #269

Issue #183, opened 2019-12-27
Quorum: Ansible roles to create namespace and serviceaccounts


sownak:
### Description
As a developer I want to create namespaces and service accounts so that I can deploy Quorum network on Kubernetes using BAF.
For each organization, create namespace and service accounts similar to that for Corda

### Acceptance Criteria
1. Get all mandatory parameters from network.yaml
2. Use Gitops to sync with Kubernetes cluster

sownak:
Closed as per #264

Issue #178, opened 2019-12-27
Quorum: Ansible role to create AWS storage class


sownak:
### Description
As a developer I want to create a storage class on AWS automatically so that the kubernetes pods can use storage provided by AWS
For reference, this role will be same as the role for Corda.
### Acceptance Criteria
1. New storage class template for AWS for Quorum
2. Check and create storage class on each cluster.

sownak:
Closed as per #264

PR #264, opened 2020-01-24
Configuring vault access to k8s

deepakkumardbd:
#178 #184 #183


platforms/quorum/configuration/roles/create/k8_component/templates/eks_storageclass.tpl
```diff
@@ -0,0 +1,9 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: {{ component_name }}
+provisioner: kubernetes.io/aws-ebs
+reclaimPolicy: Delete
+volumeBindingMode: Immediate
+parameters:
+  encrypted: "true"
```
sownak: please add newline in all files. This is a github requirement

platforms/quorum/configuration/roles/create/k8_component/templates/eks_storageclass.tpl
```diff
@@ -0,0 +1,10 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: {{ component_name }}
+provisioner: kubernetes.io/aws-ebs
+reclaimPolicy: Delete
+volumeBindingMode: Immediate
+parameters:
+  encrypted: "true" 
+
```
sownak: This is not right, the red mark there should not be there if you did it correctly. For all the files.

Issue #180, opened 2019-12-27
Quorum: Ansible role to create crypto for RAFT


sownak:
### Description
As a developer I want to create crypto materials for RAFT so that I can deploy Quorum using BAF with RAFT consensus.
#### Pre-requisites
Add two new roles to install geth and bootnode.
Steps:
1. Check if geth/bootnode is already installed, if yes, skip.
2. If no, then from build directory `git clone https://github.com/jpmorganchase/quorum.git`
3.  From `quorum` directory, execute `make geth` and `make bootnode`, depending on which binary you are installing.
4. Move geth and/or bootnode binary from quorum/build/bin/ to defined bin directory.

#### Steps for RAFT crypto
1. For each org, first check Vault if the nodekey crypto exists. If exists, skip all following steps. 
2. If not, then create a node directory `mkdir <node name>`
3. Then `bootnode -genkey <node name>/nodekey`
4. Step to generate enode URL for this node (to add in static-nodes.json) `bootnode --nodekey=<node name>/nodekey --writeaddress > <node name>/enode`
5. Step to create a new geth account and keystore
  a) create a file called password to store the passphrase `echo {{peer.geth_passphrase}} >> password`
  b) `geth account new --datadir <node name> --password password`
  c) `cp <node name>/keystore/* <node name>/keystore.json`
6. Put Crypto in Vault
`vault kv put secret/<namespace>/crypto/<node name>/quorum nodekey="$(cat <node name>/nodekey)" keystore="$(cat <node name>/keystore.json)" db_user="demouser" db_password="password" gethpassword="$(cat password)"`

### Acceptance Criteria
1. All pre-requisites installed for RAFT crypto generation.
2. Crypto generation is idempotent i.e. generate new crypto if it does not exist.
3. Crypto is visible in Vault at path secret/<namespace>/crypto/<peername>/quorum.

jagpreetsinghsasan:
![MicrosoftTeams-image (2)](https://user-images.githubusercontent.com/56861721/73067877-c7d50400-3ecf-11ea-92d2-ea93e1e0bb73.png)

sownak:
Closed as per #268

PR #268, opened 2020-01-24
Quorum crypto generation for RAFT consensus

jagpreetsinghsasan:
#180

### DIFF
--- a/platforms/quorum/configuration/roles/setup/bootnode/vars/main.yaml
+++ b/platforms/quorum/configuration/roles/setup/bootnode/vars/main.yaml
@@ -4,3 +4,4 @@ tmp_directory: "{{ lookup('env', 'TMPDIR') | default('/tmp',true) }}"
 bootnode:
   repo: https://github.com/jpmorganchase/quorum.git
   bin_directory: ~/bin
+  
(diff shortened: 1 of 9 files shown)


platforms/quorum/configuration/roles/setup/geth/tasks/main.yaml
```diff
@@ -0,0 +1,43 @@
...
+- name: Make geth
+  shell: |
+    cd "{{ tmp_directory.path }}"
+    make geth
+  when: geth_stat_result.stat.exists == False
+
+# This task creates the bin directory, if it doesn't exist, for storing the geth binary
+- name: Create bin directory
+  file:
+    path: "{{ geth.bin_directory }}"
```
sownak: Please use the global bin directory (from the playbook)

platforms/quorum/configuration/roles/setup/geth/tasks/main.yaml
```diff
@@ -0,0 +1,43 @@
...
+  when: geth_stat_result.stat.exists == False
+
+# This task puts the geth binary to above created bin directory
+- name: Copy geth binary to destination directory
+  copy:
+    src: "{{ tmp_directory.path }}/build/bin/geth"
+    dest: "{{ geth.bin_directory }}/geth"
+    mode: 0755
+    remote_src: yes
+  when: geth_stat_result.stat.exists == False
```
sownak: Add newline

platforms/quorum/configuration/roles/setup/bootnode/vars/main.yaml
```diff
@@ -0,0 +1,6 @@
+---
+tmp_directory: "{{ lookup('env', 'TMPDIR') | default('/tmp',true) }}"
+
+bootnode:
+  repo: https://github.com/jpmorganchase/quorum.git
+  bin_directory: ~/bin
```
sownak: use global bin directory. Otherwise we will be changing at multiple places

Issue #267, opened 2020-01-24
authority-baf-ac-auth-job and authority-authority-trustee-auth-job fails because kubernetes_url is empty


sownak:
**Describe the bug**
authority-baf-ac-auth-job and authority-authority-trustee-auth-job fails because kubernetes_url is empty

**To Reproduce**
1. Edit the samples/network-indyv3.yaml according to configuration
2. execute ansible-playbook site.yaml -e "samples/network-indyv3.yaml"
3. authority-baf-ac-auth-job and authority-authority-trustee-auth-job fails and goes into crashloopbackoff
4. Playbook execution fails

**Expected behavior**
The auth-jobs should be completed and playbook execution should continue.

**Additional context**
"Get the kubernetes server url" is not giving the correct output.

juraj-durech:
Closed via #307

Issue #265, opened 2020-01-24
Ansible fails at TASK [setup/vault_kubernetes : Write reviewer token for Organizations] 


sownak:
**Describe the bug**
Ansible fails at TASK [setup/vault_kubernetes : Write reviewer token for Organizations] 
This is because the kubernetes.context is not used, instead the context is assumed to be "kubernetes". This would not work if the Cluster Context is not kubernetes.

**To Reproduce**
1. Edit the samples/network-indyv3.yaml according to configuration
2. execute ansible-playbook site.yaml -e "samples/network-indyv3.yaml"
3. Execution fails when at the vault_kubernetes role.

**Expected behavior**
Execution to continue and update Vault with new key/value.

**Screenshots**
![image](https://user-images.githubusercontent.com/20537268/73064871-19fb3280-3e9a-11ea-841d-879740abbd7f.png)

juraj-durech:
Closed via #307

PR #307, opened 2020-01-30
Fix for Indy bugs and other issues

sownak:
Fixed: #267 , #265 , 
Added "executable" to run Ansible shell on bash (as the += operator fails in sh shell)
Reduced the PVC volume from 10GB to 2 GB
Added "token" in gitignore otherwise it will get checked in by other playbooks.


sownak:
Please merge using "Squash and merge" so that irrelevant commits are not added to github history

juraj-durech:
@sownak I see files from the Quorum platform being deleted in this PR, are you ok with this?

sownak:
> @sownak I see files from the Quorum platform being deleted in this PR, are you ok with this?

Yes, these charts are to be deleted as we have common node_* charts for quorum

PR #317, opened 2020-02-03
End of Sprint 7

sownak:


platforms/quorum/configuration/roles/create/generate_crypto_raft/tasks/main.yaml
```diff
@@ -0,0 +1,38 @@
+# This task checks if the crypto material is already stored in the vault or not
```
sownak: path needs to be updated

PR #323, opened 2020-02-05
permission patch

lakshyakumar:
Signed-off-by: Kumar <lakshya.kumar@accenture.com>

Adding the patch for permissions required for msp

### DIFF
--- a/platforms/hyperledger-fabric/configuration/roles/create/crypto/orderer/tasks/orderer.yaml
+++ b/platforms/hyperledger-fabric/configuration/roles/create/crypto/orderer/tasks/orderer.yaml
@@ -41,6 +41,18 @@
   when: vault_msp_result.failed == True
 
 ############################################################################################
+# This task changes the permissions for msp
+- name: Convert the file to unix format
+  shell: |
+    chmod 755 ./build/{{ files }}
+  with_items:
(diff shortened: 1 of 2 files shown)


platforms/hyperledger-fabric/configuration/roles/create/crypto/orderer/tasks/orderer.yaml
```diff
@@ -40,6 +40,18 @@
     loop_var: files
   when: vault_msp_result.failed == True
 
+############################################################################################
+# This task changes the permissions for msp
+- name: Convert the file to unix format
+  shell: |
```
sownak: use file module https://docs.ansible.com/ansible/latest/modules/file_module.html instead of shell

platforms/hyperledger-fabric/configuration/roles/create/crypto/peer/tasks/common.yaml
```diff
@@ -25,6 +25,18 @@
     loop_var: files
   when: vault_msp_result.failed == True
 
+############################################################################################
+# This task changes the permission for msp
+- name: Convert the file to unix format
```
sownak: same as above. Task name is "Change permissions of file"

Issue #179, opened 2019-12-27
Quorum: Ansible role to create crypto for IBFT


sownak:
### Description
As a developer I want to create crypto materials for IBFT so that I can deploy Quorum using BAF with IBFT consensus.

#### Pre-requisites
Add new role to install istanbul-tools.
Steps:
1. Check if istanbul is already installed, if yes, skip.
2. If no, then from build directory `git clone https://github.com/jpmorganchase/istanbul-tools.git`
3.  From `istanbul-tools` directory, execute `make`.
4. Move istanbul binary from istanbul-tools/build/bin/istanbul to defined bin directory.
5. Also install geth and bootnode if they do not exist.

#### Steps for IBFT crypto
1. Genesis.json and nodekey/enode for all orgs has to be created by a single command in IBFT, hence the following steps should be executed once per initial Quorum network.
  a) Input to this role: Number of Validator nodes from network.yaml, org_name and node_name for all validator nodes.
  b) first check Vault if the nodekey exists for nodes. If exists, skip all following steps as Genesis is already done.
  c) If not, generate istanbul files `istanbul setup --num <number of validator nodes> --nodes --quorum --save --verbose`
  d) Rename the generated directories with the node names(<node name>).
2. For each node, generate enode URL for this node (to add in static-nodes.json) `bootnode --nodekey=<node name>/nodekey --writeaddress > <node name>/enode`
3. Step to create a new geth account and keystore
  a) create a file called password to store the passphrase `echo {{peer.geth_passphrase}} >> password`
  b) `geth account new --datadir <node name> --password password`
  c) `cp <node name>/keystore/* <node name>/keystore.json`
4. Put Crypto in Vault
`vault kv put secret/<namespace>/crypto/<node name>/quorum nodekey="$(cat <node name>/nodekey)" keystore="$(cat <node name>/keystore.json)" db_user="demouser" db_password="password" gethpassword="$(cat password)"`

### Acceptance Criteria
1. All pre-requisites installed for IBFT crypto generation.
2. Crypto generation is idempotent i.e. generate new crypto if it does not exist.
3. Crypto is visible in Vault at path secret/<namespace>/crypto/<peername>/quorum.

PR #325, opened 2020-02-06
Crypto IBFT

jagpreetsinghsasan:
#179

### DIFF
--- a/platforms/quorum/configuration/roles/create/crypto/ibft/meta/main.yaml
+++ b/platforms/quorum/configuration/roles/create/crypto/ibft/meta/main.yaml
@@ -1,0 +1,6 @@
+---
+dependencies:
+  - role: "setup/geth"
+    when: network.type == 'quorum'
+  - role: "setup/bootnode"
+    when: network.type == 'quorum'
--- a/platforms/quorum/configuration/roles/create/crypto/ibft/meta/main.yaml
+++ b/platforms/quorum/configuration/roles/create/crypto/ibft/meta/main.yaml
@@ -1,6 +1,4 @@
 ---
 dependencies:
   - role: "setup/geth"
-    when: network.type == 'quorum'
   - role: "setup/bootnode"
-    when: network.type == 'quorum'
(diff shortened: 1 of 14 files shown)


platforms/quorum/configuration/roles/create/crypto/ibft/tasks/nested_main.yaml
```diff
@@ -0,0 +1,35 @@
...
+  file:
+    path: build/{{ component_name }}/{{ peer.name }}
+    state: directory
+    mode: '0755'
+    recurse: yes
+  when: vault_nodekey_result.failed == True
+
+- name: Generate enode url for each node and create a geth account and keystore
+  shell: |
+    bootnode --nodekey=build/{{ component_name }}/{{ peer.name }}/nodekey --writeaddress > build/{{ component_name }}/{{ peer.name }}/enode
```
sownak: add bootnode and geth roles in the meta, so that the dependencies are strong. That way we do not have to explicitly call bootnode and geth roles

platforms/quorum/configuration/roles/create/genesis_nodekey/tasks/main.yaml
```diff
@@ -0,0 +1,46 @@
...
+    state: directory
+    mode: '0755'
+    recurse: yes
+  when: vault_check|default(0)|int != 0
+
+# This task generates the istanbul files
+- name: Generate istanbul files
+  shell: |
+    cd build
+    istanbul setup --num {{ validator_node_list | length }} --nodes --quorum --save --verbose
```
sownak: same for this. Add in meta

platforms/quorum/configuration/roles/create/crypto/ibft/meta/main.yaml
```diff
@@ -0,0 +1,6 @@
+---
+dependencies:
+  - role: "setup/geth"
+    when: network.type == 'quorum'
```
sownak: when is not needed.

platforms/quorum/configuration/roles/create/genesis_nodekey/meta/main.yaml
```diff
@@ -0,0 +1,6 @@
+---
+dependencies:
+  - role: "setup/geth"
+    when: network.type == 'quorum'
```
sownak: no when

platforms/quorum/configuration/roles/create/crypto/ibft/meta/main.yaml
jagpreetsinghsasan: Ok

Issue #81, opened 2019-11-05
Indy: Add a network.yaml HL Indy validator Ansible role


juraj-durech:
Validate e.g.

1. At least 4 genesis stewards
2. Max 1 trustee per org
3. Max 1 endorser per org
4. At least one trustee per network.yaml
...

juraj-durech:
Closed via #332

PR #329, opened 2020-02-10
Indy

himanshulalarya:
#81


platforms/hyperledger-indy/configuration/roles/check/validation/tasks/check_count.yaml
```diff
@@ -0,0 +1,36 @@
...
+    trustee_count=0
+  
+- name: "Counting trustees per Org"
+  set_fact:
+    trustee_count={{ trustee_count|default(0)|int + 1 }}
+    total_trustees={{ total_trustees||default(0)int + 1 }}
+  loop: "{{ trustees }}"
+
+- name: Print error if trustee count limit fails
+  debug: msg="The trustee count is {{ trustee_count }}. There should be max 1 trustee per organization."
```
sownak: Will this fail the playbook? use failed_when, check https://docs.ansible.com/ansible/latest/user_guide/playbooks_error_handling.html

platforms/hyperledger-indy/configuration/roles/check/validation/tasks/main.yaml
```diff
@@ -0,0 +1,36 @@
...
+    loop_var: organizationItem
+  when: network.type == 'indy'
+
+
+- name: Print error if genesis steward count limit fails
+  debug: msg="The genesis count is {{ steward_count }}. There should be atleast 4 genesis steward in total."
+  when: steward_count|int < 4
+
+- name: End Play if genesis endorser count is less than 4
+  meta: end_play
```
sownak: same comment as previous. use failed_when

platforms/quorum/charts/raft_constellation/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart to deploy Quorum nodes with RAFT consensus and constellation transaction manager
```
sownak: You need to reset from upstream:develop these files/chats were removed.

platforms/hyperledger-indy/configuration/roles/check/validation/tasks/check_count.yaml
himanshulalarya: Yes, if the condition is unsatisfied , it will fail and the playbook execution will be stopped. It will give the error msg and will not proceed any furthur.

Issue #298, opened 2020-01-29
Quorum App: Update Product Custodian


sownak:
### Description
As a Participant in the Blockchain network I want to update  the current custodian of a product when it changes hands so that I can track the change of custodian.
Implement the transaction to update a Products custodian.

The transaction should accept a Product ID and a string consisting of Lat Long coordinates.

Then update the custodian of the given Product to our Identity, and the lastScannedAt field to the supplied string.

Given that:
The Product exists, else return HTTP 404
Our Identity is in the participants list, else return HTTP 404
The Product's containerID field is "" (empty string) otherwise return HTTP 400
### Acceptance Criteria
1. Follow Solidity coding standards
2. Product will have a new custodian

PR #343, opened 2020-02-14
Product smart contract

Rayas-NG:
added updateCustodian function for issue #298


sownak:
Please update the PR with target branch as develop and not master

PR #345, opened 2020-02-14
Quorum api get all products

jane-b-lee:
added getAllProducts() function in the smart contract, Express/web3 call function to return all added products.


sownak:
@jane-b-lee  there appears to be conflict with this PR.

PR #364, opened 2020-02-17
#Fix for 188

deepakkumardbd:
Signed-off-by: Kumar, D. bd <d.bd.kumar@accenture.com>

### DIFF
--- a/platforms/quorum/configuration/roles/helm_component/templates/constellation.tpl
+++ b/platforms/quorum/configuration/roles/helm_component/templates/constellation.tpl
@@ -1,0 +1,63 @@
+apiVersion: flux.weave.works/v1beta1
+kind: HelmRelease
+metadata:
+  name: {{ component_name }}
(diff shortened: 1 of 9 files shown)


platforms/quorum/configuration/roles/helm_component/templates/constellation.tpl
```diff
@@ -0,0 +1,63 @@
...
+      storage: "bdb:/etc/quorum/qdata/database"
+      tls: {{ network.config.tm_tls }}
+      othernodes: {{ network.config.tm_nodes }}
+      trust: {{ network.config.tm_trust }}
+    proxy:
+      provider: "ambassador"
+      external_url_suffix: {{ external_url }}
+      rpcport: {{ peer.rpc.ambassador }}
+      quorumport: {{ peer.p2p.ambassador }}
+      portConst: 1503
```
sownak: This should be from the network.yaml

Issue #189, opened 2019-12-27
Quorum: Ansible playbook for resetting network


sownak:
### Description
As a developer I want to create an ansible-playbook so that I can delete a Quorum network with the given configuration file.

### Acceptance Criteria
1. All prerequisites completed for the network
2. All components of the Quorum network should be deleted includes vault keys and namespaces.

PR #393, opened 2020-02-25
Reset network playbook for Quorum

jagpreetsinghsasan:
#189

### DIFF
--- a/platforms/quorum/configuration/reset-network.yaml
+++ b/platforms/quorum/configuration/reset-network.yaml
@@ -1,0 +1,52 @@
+# This playbook deletes the DLT network on existing Kubernetes clusters which has been created
+#  using the Blockchain Automation Framework.
+# Please use the same network.yaml to run this playbook as used for deploy-network.yaml
+###########################################################################################
+# THIS PLAYBOOK DELETES EVERYTHING, EVEN NAMESPACES and FLUX
+###########################################################################################
+---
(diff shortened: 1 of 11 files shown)


platforms/quorum/configuration/reset-network.yaml
```diff
@@ -0,0 +1,52 @@
...
+    # delete build directory
+    - name: Remove build directory
+      file:
+        path: "./build"
+        state: absent
+  vars: #These variables can be overriden from the command line
+    privilege_escalate: false           #Default to NOT escalate to root privledges
+    install_os: "linux"                 #Default to linux OS
+    install_arch:  "amd64"              #Default to amd64 architecture
+    bin_install_dir:  "/bin"            #Default to /bin install directory for binaries
```
sownak: use ~/bin

platforms/quorum/configuration/roles/delete/vault_secrets/tasks/main.yaml
```diff
@@ -0,0 +1,50 @@
+#############################################################################################
+# This role deletes the Vault configurations
+#############################################################################################
+
+#############################################################################################
+# This task deletes the Docker credentials
+- name: Delete docker creds
+  k8s_raw:
```
sownak: does this work with k8s? then use the latest module

PR #401, opened 2020-02-26
Quorum

himanshulalarya:
#349


docs/source/developer/quorum-ansible.md
```diff
@@ -0,0 +1,243 @@
...
+## **create/tessera**
+
+* Set enode_data_list to []
+* Get enode data for each node of all organization
+* Get enode data
+* Get enode_data
+* Get validator node data
+
+Follow [Readme](https://github.com/hyperledger-labs/blockchain-automation-framework/tree/master/platforms/quorum/configuration/roles/create/tessera/Readme.md) for detailed information.
+
```
sownak: Add the "delete" roles as well, this is available on develop now. Also include reset_network

himanshulalarya: Added

PR #414, opened 2020-02-26
Fabric improvements

Henni:
Various improvements and fixes for the Hyperledger Fabric codebase:

- fix `create-join-channel.yaml` playbook
- clearer naming for `peer/crypto` tasks
- typo fix: invokation -> invocation
- fix `reset-network.yaml` to uninstall flux before removing helmreleases
- remove redundant network.type checks (`network.type == 'fabric'`) within Fabric playbooks
- fix `generate-artefacts-deploy.yaml` playbook


Henni:
If you prefer these minor changes to be split up into their own PRs respectively, please let me know.

platforms/hyperledger-fabric/configuration/reset-network.yaml
```diff
@@ -39,7 +39,15 @@
         gitops: "{{ item.gitops }}"
         release_dir: "{{playbook_dir}}/../../../{{item.gitops.release_dir}}/{{ item.name|lower }}"
       loop: "{{ network['organizations'] }}"
-      when: network.type == 'fabric'
+
+    # Uninstalling Flux is needed so that everything is clean
+    # remove this if not needed
+    - name: Uninstall flux
```
sownak: This should be in delete/flux_releases because then it works per org, and our assumption is that each org is a separate cluster i.e will have different flux deployed.

platforms/hyperledger-fabric/configuration/roles/delete/flux_releases/tasks/main.yaml
```diff
@@ -66,14 +66,6 @@
   when: component_type == 'peer'
   ignore_errors: yes
 
-# Uninstalling Flux is needed so that everything is clean
-# remove this if not needed
```
sownak: as per the prev comment, this should stay.

platforms/hyperledger-fabric/configuration/reset-network.yaml
Henni: The issue with keeping this in `delete/flux_releases` is that it deletes flux before some helmreleases if there are multiple orgs deployed to a single cluster.
I believe this prevents proper cleanup of the network.

Do you have an idea how to fix this properly?

sownak: In that case, we need a separate role to delete just the flux and run that in a loop, so this code works, just it needs to run in network.orgs loop.

Henni:
@jagpreetsinghsasan pulling you in for the discussion on bf1a1b0

You mentioned that playbook conditionals won't work here.
But the article you mentioned confirms that variables passed through `--extra-vars`should work fine (which I was able to confirm myself as well).
Could you please double check this is an issue for you or tell me how you pass the network.yaml to Ansible.

Henni:
Found the actual issue with the reset-network playbook:
It wasn't about uninstalling flux multiple times. The actual issue was that flux was uninstalled after the helm releases where removed. In between these steps flux sometimes reinstalled some of the releases. Fixed in 0d9822b

PR #446, opened 2020-03-11
Update docs, gitignore and add missing 'check' Ansible role for Quorum

uyenle57:
This PR includes:

- Update typos, cleanup syntax, and add new info to documentations.
- Add missing 'check' Ansible role to `platforms/quorum/configuration/roles`
- Update .gitignore to include Vault configs that users would create while following tutorials but should not be version controlled.


docs/source/developer/baf_minikube_setup.md
```diff
@@ -73,36 +88,75 @@
...
-      username: "<github_username>"          
-      password: "<github_token>"         
-      email: "<github_email>"
+     git_ssh: <ssh url of your forked repo>
+     branch: "develop"
+     git_push_url: <https url of your forked repo without the https://)
+     username: "<github_username>"
+     password: "<github token/password>"
+     email: "<github_email>"
+     private_key: "/home/blockchain-automation-framework/build/gitops.pem"
```
sownak: "gitops" only, should be hardcoded in the network.yaml sample hence not required here.

platforms/quorum/releases/dev/carrier-ns/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
sownak: do not check-in contents of releases folder. Remove "dev" before each pull request

vault/core/_audit
```diff
@@ -0,0 +1 @@
+{"Value":"AAAAAQLqm3nyg2QzB/a+ZnYay7oGLXCUkDHOoH6pV+EtV2PNzl3RdgZevQQaFZPgxwsz0kaVpMMI0/PWJN54+li2xVWHzFLEr6TAFWQh94DVo4D6GKcIeExh"}
```
sownak: ideally, vault should be outside blockchain-automation-framework, but you have now added gitignore

sownak:
I have added a few changes to the dev guide, so refresh your branch as well.

Issue #46, opened 2019-10-29
Implement RAFT Consensus for Fabric


sownak:
As a user I want to use RAFT Consensus for my Fabric network so that I have a choice of consensus mechanisms.

This is a stretch goal if Fabric 1.4.1 is implemented.

### Acceptance Criteria:
 1. Add RAFT as another consensus mechanism
 2. Provide choice of consensus mechanism in network configuration.
 3. Create or update charts as necessary.
 4. Test deployment on cluster using GitOps.

sownak:
Priority.

jagpreetsinghsasan:
The current supplychain chaincode doesnt work.
So I installed fabcar chaincode to test the network.
I am attaching the screenshots of the query regarding the same

jagpreetsinghsasan:
![alpha](https://user-images.githubusercontent.com/56861721/77397179-1d903480-6dcb-11ea-90ac-915ad4727ee1.PNG)
![beta](https://user-images.githubusercontent.com/56861721/77397187-21bc5200-6dcb-11ea-9657-afaf0d2c32ec.jpg)

jagpreetsinghsasan:
To validate network I performed the following steps
1. Setup Fabric network using BAF and RAFT
2. Install 2 peer cli, one of each peer of the org (There are 2 orgs in my network.yaml with one peer in each org)
3. For peer0 of org1, perform the following steps in its cli
curl https://raw.githubusercontent.com/hyperledger/fabric-samples/release-1.4/chaincode/fabcar/go/fabcar.go > fabcar.go
mkdir -p /opt/gopath/src/github.com/chaincode
cp fabcar.go /opt/gopath/src/github.com/chaincode
peer chaincode install -n fabcar -v 1.0 -p github.com/chaincode

vault read secret/crypto/ordererOrganizations/sj-net/orderers/orderer3.sj-net/tls (get ca.crt) and store it as orderer.pem

peer chaincode instantiate -o orderer3.hf.dev.aws.blockchaincloudpoc-develop.com:8443 --tls --cafile /opt/gopath/src/github.com/chaincode/orderer.pem -C allchannel -n fabcar -v 1.0 -c '{"Args":["Init"]}' -P ""

vault read secret/crypto/peerOrganizations/cj-net/peers/peer0.cj-net/tls   (get ca.crt) and store it as peer.pem

[...]

PR #490, opened 2020-03-24
Fabricraft merge

jagpreetsinghsasan:
#46


platforms/hyperledger-fabric/charts/orderernode/templates/deployment.yaml
```diff
@@ -52,6 +52,8 @@ spec:
         args:
         - |-
           #!/usr/bin/env sh
+          if [ {{ $.Values.consensus.name }} == kafka ]
```
sownak: Shouldn't this have ''?

platforms/hyperledger-fabric/configuration/roles/create/configtx/templates/configtxOrderer.tpl
```diff
@@ -2,11 +2,11 @@ Orderer: &OrdererDefaults
   OrdererType: {{ consensus.name }}
   Addresses:
 {% for orderer in orderers %}
-  {%- if provider == 'minikube' %}
+{% if provider == 'minikube' %}
```
sownak: removing - will keep newlines. Is this needed?

docs/source/operations/fabric_networkyaml.md
```diff
@@ -396,7 +396,7 @@ The fields under `orderer` service are
 |-------------|----------------------------------------------------------|
 | name                        | Name of the Orderer service                                                                                                     |
 | type          | This type must be `orderer`  |
-| consensus                   | Consensus type, for example: kafka                                                                               |
+| consensus                   | Consensus type, for example: kafka, etcdraft                                                                               |
```
sownak: call it "raft" like we do in quorum

platforms/hyperledger-fabric/configuration/roles/create/configtx/templates/configtxOrderer.tpl
jagpreetsinghsasan: the list looks better with new lines
earlier it was like 
addresses:
  - a.com   - b.com   -c.com

now it is 
addresses:
  - a.com
  - b.com
  - c.com

platforms/hyperledger-fabric/charts/orderernode/templates/deployment.yaml
jagpreetsinghsasan: I dont think we need quotes (as it is not the first character of the string.
I have tested both kafka and raft and both seems fine with this

docs/source/operations/fabric_networkyaml.md
jagpreetsinghsasan: sure. I will do the changes

Issue #485, opened 2020-03-20
Fabric 1.4.4 Test on Fabric 2.0 - Error on Create Channels


himanshulalarya:
**Describe the bug**
While running the feature/Fabric 2.0 network with fabric 1.4.4 network.yaml file , it is generating implicit policy error at create channels

PR #504, opened 2020-03-30
Fabric test

himanshulalarya:
Bug Fix For create channels
#485


platforms/hyperledger-fabric/configuration/roles/create/channel_artifacts/tasks/main.yaml
```diff
@@ -55,7 +55,11 @@
 - name: "Creating genesis block"
   shell: |
     cd {{ build_path }}
+    {% if network.version == '2.0' %}
```
sownak: use "contains" or similar, as this will be applicable to 2.x.x

platforms/hyperledger-fabric/configuration/roles/create/channel_artifacts/tasks/main.yaml
```diff
@@ -55,7 +55,11 @@
...
     ./configtxgen -profile {{ genesis.name }} -channelID {{ channel_name }}  -outputBlock ./channel-artifacts/genesis.block
```
sownak: If this is per channel, then we have to rethink when the genesis block is generated (or force only one channel when the network is in genesis mode)

platforms/hyperledger-fabric/configuration/roles/create/configtx/templates/configtxOrderer.tpl
```diff
@@ -22,6 +22,7 @@ Orderer: &OrdererDefaults
 {% endfor %}
 {% endif %}
   Organizations:
+{% if network.version == '2.0' %}
```
sownak: same comment as previous

platforms/hyperledger-fabric/configuration/roles/create/configtx/templates/configtxOrg.tpl
```diff
@@ -2,6 +2,7 @@
     Name: {{ component_name }}MSP
     ID: {{ component_name }}MSP
     MSPDir: ./crypto-config/{{ component_type }}Organizations/{{ component_ns }}/msp
+{% if network.version == '2.0' %}
```
sownak: change here too

platforms/hyperledger-fabric/configuration/roles/create/configtx/templates/configtxinit.tpl
```diff
@@ -9,6 +9,7 @@ Capabilities:
 
 Application: &ApplicationDefaults
   Organizations:
+{% if network.version == '2.0' %}
```
sownak: same comment

platforms/hyperledger-fabric/configuration/roles/create/channel_artifacts/tasks/main.yaml
himanshulalarya: Genesis block is per channel only. So it should be fine as for a particular genesis block we will be passing its channel only

Issue #539, opened 2020-03-31
Custodian should contain the Node OU as well


sownak:
**Describe the bug**
In all the responses from the API, the product/container custodian should contain OU like "O=Manufacturer,OU=Manufacturer,L=47.38/8.54/Zurich,C=CH" for the respective address. 

The Node Identity can be fetched from web3.admin.nodeInfo.name and then this should be appended to the custodian along with node/eth address

**To Reproduce**
1. Create a Quorum network with IBFT + tessera
2. Deploy general smartcontract on it
3. Start expressAPI and create a product
4. Get the product and check the JSON response.

**Expected behavior**
The custodian should have the value like "0x6334c91827ca290ac93b4ecbcba2d37e136f5462,O=Manufacturer,OU=Manufacturer,L=47.38/8.54/Zurich,C=CH" when the nodeInfo.name is "Geth/O=Manufacturer,OU=Manufacturer,L=47.38/8.54/Zurich,C=CH/v1.8.18-stable-685f59fb(quorum-v2.5.0)/linux-amd64/go1.11.13"

Issue #538, opened 2020-03-31
Quorum API scan is not returning correct value


sownak:
**Describe the bug**
Quorum API :productID/scan and :containerID/scan is not returning correct value. 

**To Reproduce**
1. Deploy Quorum network using BAF (IBFT + tessera)
2. Deploy smartcontract on the network
3. Start Quorum express-api server
4. Call :productID/scan
5. response is following
{
	"status": {
		"blockHash": "0x76dfa7388075ac6a4c3a6c9831289c16adcb1394a692877e1bf2e3e5f37726c1",
		"blockNumber": 582,
		"contractAddress": null,
		"cumulativeGasUsed": 28600,
		"from": "0xe219decb02cb1ee765cb8aad429c07edd5f4092a",
		"gasUsed": 28600,
		"logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
		"status": true,
		"to": "0xc45d5238d5480be29e5651d1a3fae7e7481acff8",
		"transactionHash": "0xed67aacaae82cd7e9d1973ade29f54b90a4b51f4f4e0f689b800bdda968db2fe",
		"transactionIndex": 0,
		"events": {}
	}
}

**Expected behavior**
Response should be like 
{
"owned"/"new"/"unowned"
}

**Additional context**
The comparison for address and custodian should be using contains and ignorecase comparison rather than a .equals or ==

PR #540, opened 2020-04-01
fix API bugs and node identity details like OU 

Rayas-NG:
issue #538 & #539


examples/supplychain-app/quorum/smartContracts/contracts/Permission.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity >0.6.0;
+pragma solidity 0.6.1;
```
sownak, line 1: use the >0.6.0 otherwise there is compiler error when solidity version is anything else

examples/supplychain-app/supplychain-frontend/package.json
```diff
@@ -32,8 +32,8 @@
   },
   "nodes": {
     "PartyA": "http://localhost:8080",
-    "PartyB": "http://localhost:8082",
-    "PartyC": "http://localhost:8084"
+    "PartyB": "http://localhost:8001",
```
sownak, line 35: revert this as well. the ports should be 8082 and 8084

Issue #505, opened 2020-03-30
Support for Quorum 2.5.0


sownak:
### Description
As a user I want to deploy Quorum 2.5.0 so that run latest quorum network and deploy smartcontracts.

### Acceptance Criteria
1. Quorum network should be up when network.version = 2.5.0
2. quorumengineering/quorum:2.5.0 container should be used
3. Quorum configurations should be updated according to latest version of geth.

PR #547, opened 2020-04-03
quorum updation

deepakkumardbd:
#505

### DIFF
--- a/examples/supplychain-app/quorum/smartContracts/truffle-config.js
+++ b/examples/supplychain-app/quorum/smartContracts/truffle-config.js
@@ -53,7 +53,7 @@ module.exports = {
       timeoutBlocks: 200
     },
     manufacturer: {
-      host: "manufacturerkd.rc.dev.aws.blockchaincloudpoc-develop.com",     // Localhost (default: none)
+      host: "manufacturer.quo.demo.aws.blockchaincloudpoc.com",     // Localhost (default: none)
       port: 15021,            // Standard Ethereum port (default: none)
       network_id: "*", // Match any network id
       gasPrice: 0,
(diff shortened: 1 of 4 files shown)


platforms/quorum/configuration/roles/helm_component/templates/constellation.tpl
```diff
@@ -54,7 +54,11 @@ spec:
 {% endfor %}
 {% endif %}
     constellation:
+{% if network.config.tm_tls == 'strict' %}
+      url: https://{{ peer.name }}.{{ external_url }}:{{ peer.transaction_manager.ambassador }}/
```
sownak, line 58: does this work?

platforms/quorum/charts/node_constellation/templates/deployment.yaml
```diff
@@ -169,8 +169,8 @@ spec:
         - "-cx"
         - |-
           #!/usr/bin/env sh
-          sleep 5;
-          
+          apk add curl;
+          # sleep 5;
```
sownak, line 173: remove commented code

examples/supplychain-app/quorum/smartContracts/truffle-config.js
```diff
@@ -53,7 +53,7 @@ module.exports = {
       timeoutBlocks: 200
     },
     manufacturer: {
-      host: "manufacturer.quo.demo.aws.blockchaincloudpoc.com",     // Localhost (default: none)
+      host: "manufacturerkd.rc.dev.aws.blockchaincloudpoc-develop.com",     // Localhost (default: none)
```
sownak, line 56: dont check in these changes. You can comment the Web3 though.

platforms/quorum/configuration/roles/helm_component/templates/constellation.tpl
deepakkumardbd: yes, in netwok.yaml we are specifying http for tm_nodes that goes to othernodes placeholder and I think that's why it is working.

Issue #523, opened 2020-03-30
Improving the dev-prerequisite automation script 


jagpreetsinghsasan:
**Describe the bug**
Currently the script doesn't have control over existing versions the software already present in the machine.  

**Expected behavior**
Script should have consistent behavior across varied configured machines 

**Environment**
 - OS: Windows

PR #561, opened 2020-04-07
Dev prerequisite update

jagpreetsinghsasan:
#523


platforms/shared/scripts/pre_setup_win.bat
```diff
@@ -37,6 +37,51 @@ PAUSE
 EXIT
 )
 
+REM ############################################################################################################################################################
+REM ############################################################################################################################################################
+
+REM Check for software versions, if present
+REM If version mismatch, prompt to delete software and exit
+SET PATH=%PATH%;%cd%\project\bin
```
sownak, line 45: \project\bin path should be first, so that it gets priority.

jagpreetsinghsasan:
> Just change the PATH. Is it possible to check versions like > rather than equals to the exact version?

I dont think we shall do this. because we dont know if the prev or upcoming versions will not break anything. We should only check for the versions on which we have tested.

Issue #465, opened 2020-03-13
Package Quorum Supplychain express_api as Docker container


sownak:
### Description
As a developer I want to run Quorum Supplychain express_api via Docker so that I can deploy it on Kubernetes.

### Acceptance Criteria
1. Express API is packages and run as Docker container similar to Corda and Fabric.
2. Docker container should connect to the smartcontract container via URLs.
3. Readme updated.

PR #580, opened 2020-04-14
Static test

Rayas-NG:
update docker file for issue #465 
added docker ignore file


sownak:
These changes are not needed. The docker-image works when the RPC server can be reached.

Issue #384, opened 2020-02-23
Enable support for dynamic addition of new node in the Quorum network using RAFT


suvajit-sarkar:
To add new nodes in RAFT mode, static-nodes json configuration file for all existing nodes, needs to be updated with new node Enode URL and then all nodes need to be restarted.

PR #595, opened 2020-04-17
Quorum newnode merge

jagpreetsinghsasan:
#384


docs/source/operations/quorum_networkyaml.md
```diff
@@ -132,6 +142,7 @@ The fields under `config` are
 | tm_nodes | The Transaction Manager nodes public addresses should be provided. For `tessera`, all participating nodes should be provided, for `constellation`, only one bootnode should be provided. NOTE The difference in the addresses for Tessera and Constellation. |
 | staticnodes | *** Existing network's static nodes need to be provided as an array. This will be implemented with addition of new node. |
 | genesis | *** Existing network's genesis.json needs to be provided in base64. This will be implemented with addition of new node.|
+| geth | *** Geth account details . This will be implemented with addition of new node.|
```
sownak, line 145: "This will be implemented with .." This text should not be replaced with actual implementation. You need to provide details of each parameter under geth. Also I think the name "geth" is a bit confusing for this parameter, change it to "bootnode"

platforms/quorum/charts/node_constellation/templates/deployment.yaml
```diff
@@ -189,8 +189,15 @@ spec:
           echo slept for 5 to give transaction manager time to start;
 
           args=""
+          NODE_STATUS=""
           if [ $CONSENSUS == 'raft' ]; then
-            args=" --raft --raftport {{ .Values.node.ports.raft }} --rpcapi admin,db,eth,debug,miner,net,shh,txpool,personal,web3,quorum,raft "
+            NODE_STATUS={{ .Values.node.status }}
```
sownak, line 194: Ensure that this is removed when ibft code is implemented. It should rather be if {{ .Values.node.status}} is defined then NODE_STATUS={{ .Values.node.status}}

platforms/quorum/charts/node_constellation/templates/deployment.yaml
```diff
@@ -189,8 +189,15 @@ spec:
...
+          fi;
+          if [ $CONSENSUS == 'raft' ] && [ $NODE_STATUS == 'new' ]; then
+            args="--raftdnsenable --raft --raftport {{ .Values.node.ports.raft }} --rpcapi admin,db,eth,debug,miner,net,shh,txpool,personal,web3,quorum,raft --raftjoinexisting {{ .Values.node.peer_id }}"
+          fi;
+          if [ $CONSENSUS == 'raft' ] && [ $NODE_STATUS == 'existing' ]; then
```
sownak, line 199: the node.status should be "new" (default) and "additional" (when new node is added in existing network). Default cannot be "existing" when we create a network from scratch.

platforms/quorum/charts/node_tessera/templates/deployment.yaml
```diff
@@ -286,8 +286,15 @@ spec:
           echo transaction manager is up;
 
           args=""
+          NODE_STATUS=""
           if [ $CONSENSUS == 'raft' ]; then
-            args=" --raft --raftport {{ .Values.node.ports.raft }} --rpcapi admin,db,eth,debug,miner,net,shh,txpool,personal,web3,quorum,raft ";
+            NODE_STATUS={{ .Values.node.status }}
+          fi;
```
sownak, line 292: same comments as for constellation

platforms/quorum/configuration/deploy-network.yaml
```diff
@@ -140,8 +141,9 @@
       docker_url: "{{ network.docker.url }}"
       charts_dir: "{{ item.gitops.chart_source }}"
       values_dir: "{{playbook_dir}}/../../../{{item.gitops.release_dir}}/{{ item.name | lower }}"
+      geth_data: "{{ network.config.geth }}"
```
sownak, line 144: change to config.bootnode as per previous comment

platforms/quorum/configuration/deploy-network.yaml
```diff
@@ -161,8 +163,52 @@
...
     loop: "{{ network['organizations'] }}"
-    when: network.config.transaction_manager == 'constellation'
+    when: network.config.transaction_manager == 'constellation' and network.config.consensus == 'raft'
+
+  # This role deploy nodes with Tessera transaction manager
+  - name: "Deploy Tessera Transaction Manager"
```
sownak, line 171: Change the task names accordingly i.e. add Tessera with Raft/IBFT

platforms/quorum/configuration/roles/create/constellation/tasks/main.yaml
```diff
@@ -3,13 +3,42 @@
   set_fact:
     enode_data_list: []
 
-# Get enode data
-- name: Get enode data for each node of the organization
+# Set node status
+- name: Set node status
+  set_fact:
+    node_status: existing
+  when: network.config.genesis | default('', true) | trim == ''
```
sownak, line 10: should rather be based on "bootnode"(geth)

platforms/quorum/configuration/roles/create/tessera/tasks/main.yaml
```diff
@@ -3,13 +3,41 @@
...
 - name: Get enode data for each node of all organization
   include_tasks: enode_data.yaml
   loop: "{{ network['organizations'] }}"
   loop_control:
     loop_var: org
 
+# This task adds a new node to the existing network
+- name: Adds a new node to the existing network
+  shell: |
+    curl -X POST http://{{ geth_data.url }}:{{ geth_data.rpcport }} --data '{"jsonrpc":"2.0","method":"raft_addPeer","params": ["enode://{{ enode.enodeval }}@{{ enode.peer_name }}.{{ external_url }}:{{ enode.p2p_ambassador }}?discport=0&raftport={{ enode.raft_ambassador }}"], "id":{{ geth_data.nodeid }} }' --header "Content-Type: application/json"
```
sownak, line 28: same comments as constellation

platforms/quorum/configuration/roles/helm_component/templates/constellation.tpl
```diff
@@ -22,6 +22,12 @@ spec:
       constellation: quorumengineering/constellation:{{ network.config.tm_version }}
     node:
       name: {{ peer.name }}
+{% if network.config.genesis | default('', true) | trim != '' %}
+{% if network.config.consensus == 'raft' %}
```
sownak, line 26: dont think the 2nd if is necessary if you dont use it in IBFT

platforms/quorum/configuration/roles/helm_component/templates/constellation.tpl
```diff
@@ -49,9 +55,16 @@ spec:
 {% endfor %}
 {% endif %}
 {% if network.config.consensus == 'raft' %}
+{% if network.config.genesis | default('', true) | trim == '' %}
```
sownak, line 58: change to bootnode as per previous comment

platforms/quorum/configuration/roles/helm_component/templates/tessera.tpl
```diff
@@ -24,6 +24,12 @@ spec:
       mysql: mysql/mysql-server:5.7
     node:
       name: {{ peer.name }}
+{% if network.config.genesis | default('', true) | trim != '' %}
+{% if network.config.consensus == 'raft' %}
```
sownak, line 28: same comments as constellation

platforms/quorum/configuration/roles/create/constellation/tasks/main.yaml
jagpreetsinghsasan: The same check can be used as it is for ibft too.
Also, genesis block information will only be available in network.yaml if there is already a network in place and we are trying to add a new org.

platforms/quorum/charts/node_constellation/templates/deployment.yaml
jagpreetsinghsasan: This will work with IBFT too as above line 193, I am setting this var to an empty string.
Earlier I was adding it in env vars, that would have failed if raft was not used. But this will do fine.

platforms/quorum/charts/node_constellation/templates/deployment.yaml
jagpreetsinghsasan: I will change this. Default and existing are better choices for the variable

platforms/quorum/configuration/roles/helm_component/templates/constellation.tpl
jagpreetsinghsasan: Second check is necessary because the peer_id is not required for IBFT.

platforms/quorum/configuration/roles/helm_component/templates/constellation.tpl
jagpreetsinghsasan: I have kept the check on genesis because its the same for IBFT and RAFT. IBFT wont have the bootnode part, but will still have the genesis part, so its better to have same checks for both.

sownak: Ok, then in the sample files, remove the genesis part in the main networkyaml, and add a new sample for add-org which has the genesis part. Otherwise users will use the main sample and then code will not work as there is genesis part in it (in case they put some bogus value in that)

jagpreetsinghsasan: Sure. That's better

platforms/quorum/configuration/roles/helm_component/templates/constellation.tpl
sownak: okay

PR #642, opened 2020-04-27
Fetching node identity and contract address from vault

suvajit-sarkar:


examples/supplychain-app/charts/expressapp-quorum/values.yaml
```diff
@@ -16,7 +16,15 @@ deployment:
   #   annotations:
   #     key: "value"
   annotations: {}
-
+vault:
```
sownak, line 19: Add comments and example for each key here.

examples/supplychain-app/charts/expressapp-quorum/values.yaml
```diff
@@ -34,9 +42,6 @@ expressapp:
   # The environment variables that will store the port to be working on and the end point to ask for requests
   #Eg. apiUrl: api.blockchaincloudpoc.com
   env:
-    ganache_server: http://manufacturer.test.quorum.blockchaincloudpoc.com:15021
```
sownak, line 37: ganache_server needs to be there, just the name should be geth_address.

examples/supplychain-app/quorum/express_nodeJS/Dockerfile
```diff
@@ -22,4 +22,3 @@ ENV NODE_ORGANIZATION="Manufacturer"
 ENV NODE_ORGANIZATIONUNIT="manufacturer"
 
 EXPOSE ${PORT}
-CMD [ "node", "app.js" ]
```
sownak, line 25: no need to change this

examples/supplychain-app/charts/expressapp-quorum/values.yaml
suvajit-sarkar: ok will add that

examples/supplychain-app/charts/expressapp-quorum/values.yaml
suvajit-sarkar: ok , but what should be the env var name for it ?

Issue #581, opened 2020-04-15
timestamp is returned as null when using RAFT Consensus


sownak:
**Describe the bug**
The "timestamp" values is returned as null when RAFT consensus is used in the Quorum network. When IBFT is used, then there is correct value.

**To Reproduce**
1. Set up a Quorum network using BAF with RAFT consensus (and any transaction manager)
2. Deploy the supplychain smartcontract
3. Start the expressapi service
4. Create a product/container and then fetch the product/container, the timestamp value is null.
5. In history api as well, the time value is null.

**Expected behavior**
The timestamp value should have a datetime value which is the time when the product/container was created.

**Additional context**
This seems to be an issue because of how RAFT handles timestamp. So the smartcontract needs to be updated accordingly.
Ref: https://github.com/jpmorganchase/quorum/issues/190

suvajit-sarkar:
This is resolved

srinivasansankaran:
@suvajit-sarkar I would like to work on it.

PR #653, opened 2020-05-01
fix timestamp for raft #581

Rayas-NG:
fix timestamp


examples/supplychain-app/quorum/smartContracts/productContract.sol
```diff
@@ -0,0 +1,41 @@
+pragma solidity 0.6.1;
+
+contract productContract {
```
sownak, line 3: What is the need for this file?

examples/supplychain-app/quorum/smartContracts/truffle-config.js
```diff
@@ -53,8 +53,9 @@ module.exports = {
       timeoutBlocks: 200
     },
     manufacturer: {
-      host: "manufacturer.quo.demo.aws.blockchaincloudpoc.com",     // Localhost (default: none)
-      port: 15021,            // Standard Ethereum port (default: none)
+      host: "192.168.99.108",     // Localhost (default: none)
```
sownak, line 56: Don't check in this.

platforms/quorum/releases/dev/carrier-ns/reviewer_rbac.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
```
sownak, line 1: Delete all files from releases folder (except readme) before raising pull request. These are specific to your deployment.

Issue #623, opened 2020-04-27
Add default molecule test scenario in Corda


sownak:
### Description
As a developer I want to run default molecule test so that I can test my ansible scripts.

### Acceptance Criteria
1.Molecule tests should cover `create/storageclass` and `create/namespace_serviceaccount` and related roles
2. Verification step should be configured.
3. Tests should complete successfully

PR #659, opened 2020-05-07
Added molecule test for storageclass, namespace, serviceaccount and clusterrolebinding for Corda platform

jagpreetsinghsasan:
#623


platforms/r3-corda/configuration/molecule/default/INSTALL.rst
```diff
@@ -0,0 +1,22 @@
+*******
```
sownak, line 1: this file is not needed

platforms/r3-corda/configuration/molecule/default/verify.yml
```diff
@@ -0,0 +1,157 @@
...
+  - name: Check if the storage class value file is created
+    stat:
+      path: "{{ playbook_dir }}/../../../releases/dev/aws-storageclass/aws-storageclass.yaml"
+    register: sc_check
+
+  - name: Assert if the value file exists or not
+    assert:
+      that: sc_check.stat.exists
+
+  - name: Deploy the resource on cluster
```
sownak, line 17: this should happen in converge, and not in verify. In verify only check.

platforms/r3-corda/configuration/molecule/default/verify.yml
```diff
@@ -0,0 +1,157 @@
...
+  - name: Check if the namespace value file is created
+    stat:
+      path: "{{ playbook_dir }}/../../../releases/dev/test-ns/namespace.yaml"
+    register: ns_check
+
+  - name: Assert if the value file exists or not
+    assert:
+      that: ns_check.stat.exists
+
+  - name: Deploy the resource on cluster
```
sownak, line 47: same comment as previous

platforms/r3-corda/configuration/roles/create/namespace_serviceaccount/tasks/main.yaml
```diff
@@ -3,62 +3,63 @@
...
   register: get_namespace
 
 # Create namespace
-- name: Create namespace for {{ organisation }}
+- name: Create namespace for {{ org.name }}
   include_role:
     name: create/k8_component
   vars:
-    component_name: "{{ component_ns }}"
+    component_name: "{{ org.name }}-ns"
```
sownak, line 16: I think component_ns should be created in the top level, so if someone wants to change the namespace then they need to make minimum changes. Here they will have to change many files. Everything else is fine.

platforms/r3-corda/configuration/roles/create/storageclass/tasks/main.yaml
```diff
@@ -24,6 +28,7 @@
 - name: "Push the created deployment files to repository"
   include_role: 
     name: "{{ playbook_dir }}/../../shared/configuration/roles/git_push"
+  tags: notest
```
sownak, line 31: use tags as a list so that new tags can be added
- notest

platforms/r3-corda/configuration/molecule/default/verify.yml
```diff
@@ -0,0 +1,157 @@
...
+  - name: Check if the vault-auth serviceaccount value file is created
+    stat:
+      path: "{{ playbook_dir }}/../../../releases/dev/test-ns/vaultAuth.yaml"
+    register: va_sa_check
+
+  - name: Assert if the value file exists or not
+    assert:
+      that: va_sa_check.stat.exists
+
+  - name: Deploy the resource on cluster
```
sownak, line 77: same comment on all. Converge is when you run the test case (here trying to create the resources with the files generated), and verify is when you just check if the resources were created.

platforms/r3-corda/configuration/roles/create/namespace_serviceaccount/tasks/main.yaml
jagpreetsinghsasan: I think, the developer should instead search for '-ns' and then check, because at some places we are explicitly doing -ns to get the namespace.

platforms/r3-corda/configuration/roles/create/storageclass/tasks/main.yaml
jagpreetsinghsasan: Sure

PR #660, opened 2020-05-07
Constellation

himanshulalarya:
#640


platforms/quorum/charts/node_constellation/templates/deployment.yaml
```diff
@@ -134,6 +134,12 @@ spec:
         args:
         - "-cx"
         - "chmod 600 $QUORUM_HOME/crypto/keys/tm.key;
+
```
sownak, line 137: Please dont cat the keys here

Issue #589, opened 2020-04-16
Re-use components check for Indy


sownak:
### Description
As a developer I want to re-use components check from platforms/shared/configuration/roles/check for Indy so that I can use the check consistently and reduce lines of code.

Refer to Issue #412

### Acceptance Criteria
1. All Directory, Vault and Docker-cred checks use the common role check/setup
2. All static K8s component (namespace, storageclass bindings etc) checks use the common role check/k8s_component
3. All dynamic K8s component (Pods, jobs etc) checks use the common role check/helm_component

PR #676, opened 2020-05-14
Add checks to shared folder

deepakkumardbd:
#589


platforms/shared/configuration/roles/check/k8_component/tasks/main.yaml
```diff
@@ -41,4 +53,22 @@
   retries: "{{ network.env.retry_count}}"
   delay: 35
   until: component_data.resources|length > 0
-  when: component_type == "ServiceAccount"
+  when: (component_type == "ServiceAccount" and (type is undefined or type == 'retry'))
+
+# Get a service account token
```
sownak, line 58: commented code?

PR #682, opened 2020-05-18
Shared roles reuse

suvajit-sarkar:
**Changelog**

- Update helm lint build directory check (helm_lint)
- Update  ambassador wait check
- Update  aws cli bin directory check
- Update flux wait check
- Update ha-proxy wait check
- Make the shared check/helm_component for modular
- fix a bug in Quorum, causing deployment to fail
- remove quorum check unused code

 

**Reviewed by**
@lakshyakumar 
 

**Linked issue**
#591


platforms/shared/configuration/roles/check/helm_component/tasks/main.yaml
```diff
@@ -30,41 +30,20 @@
...
   k8s_info:
     kind: "Pod"
     namespace: "{{ namespace }}"
-    kubeconfig: "{{ kubernetes.config_file }}"
-    context: "{{ kubernetes.context }}"
-    label_selectors:
-      - app = {{ component_name }}
+    kubeconfig: "{{ kubeconfig | default(kubernetes.config_file) }}"
+    context: "{{ context | default(kubernetes.context) }}"
+    label_selectors: "{{ label_selectors }}"
```
sownak, line 42: this line is not needed as it is not transforming anything. Instead add a comment for this role and update the readme file as to how the label_selectors should be populated when calling this role.

platforms/shared/configuration/roles/helm_lint/tasks/main.yaml
```diff
@@ -14,10 +14,11 @@
 
 # create temp test directory
 - name: "Check or create test directory"
-  file:
+  include_role:
+    name: "../../shared/configuration/roles/check/setup"
```
sownak, line 18: role name can be only check/setup ; they are in the same dir

suvajit-sarkar: yes even i thought so but looks like it takes the reference from the playbook which ran it

platforms/shared/configuration/roles/check/helm_component/tasks/main.yaml
suvajit-sarkar: Yeah got it, make sense.. will update that

platforms/shared/configuration/roles/helm_lint/tasks/main.yaml
sownak: ok, then this is fine.

Issue #681, opened 2020-05-15
'type' is undefined


sownak:
**Describe the bug**
'type' is undefined when running Quorum deployment
TASK [Checking for the carrier-quo-role-tokenreview-binding] *******************

TASK [../../shared/configuration/roles/check/k8_component : Check ClusterRoleBinding carrier-quo-role-tokenreview-binding is created] ***
fatal: [localhost]: FAILED! => {"msg": "The conditional check '(component_type == \"Namespace\" and type == \"no_retry\") or ( component_type == \"StorageClass\" and type == \"no_retry\" ) or (component_type == \"ClusterRoleBinding\" and type == \"no_retry\")' failed. The error was: error while evaluating conditional ((component_type == \"Namespace\" and type == \"no_retry\") or ( component_type == \"StorageClass\" and type == \"no_retry\" ) or (component_type == \"ClusterRoleBinding\" and type == \"no_retry\")): 'type' is undefined\n\nThe error appears to be in '/jenkins/workspace/BAF_Demo_multi_quorum-GNPANTIXFAVEYAQB6ZDJIEMQ5GAEN2YJXF6UPSXDSZK7S6PPOAGQ/platforms/shared/configuration/roles/check/k8_component/tasks/main.yaml': line 7, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n# Check if component created without retry\n- name: Check {{ component_type }} {{ component_name }} is created\n  ^ here\nWe could be wrong, but this one looks like it might be an issue with\nmissing quotes. Always quote template expression brackets when they\nstart a value. For instance:\n\n    with_items:\n      - {{ foo }}\n\nShould be written as:\n\n    with_items:\n      - \"{{ foo }}\"\n"}

**To Reproduce**
1. Edit network yaml as per instructions
2. Deploy Quorum network using BAF
3. See error

**Expected behavior**
Task should not fail

suvajit-sarkar:
have fixed this with the PR #682

PR #683, opened 2020-05-18
Fix bug

lakshyakumar:
Signed-off-by: lakshyakumar <lakshya.kumar@accenture.com>

**Changelog**
- Fix the type error in cluster role binding and service account checks for fabric and quorum

 

**Reviewed by**
@jagpreetsinghsasan 

 

**Linked issue**
#681  

 

Commit message
-------------
 Fix the type error in cluster role binding and service account checks for fabric and quorum


jagpreetsinghsasan:
Code change seems fine.

Issue #513, opened 2020-03-30
Hyperledger Besu: Ansible role to create AWS storage class


suvajit-sarkar:
### Description
As a developer I want to create a storage class on AWS automatically so that the kubernetes pods can use storage provided by AWS

For reference, this role will be same as the role for Quorum.

(add more description and relevant links)

### Acceptance Criteria
1. New storage class template for AWS for Hyperledger Besu
2. Check and create storage class on each cluster.

PR #712, opened 2020-06-03
Besu

himanshulalarya:
**Changelog**
- Added Role for creating Storage Class

 

**Reviewed by**
@sownak 

 

**Linked issue**
#513


platforms/hyperledger-besu/configuration/deploy-network.yaml
```diff
@@ -25,6 +25,17 @@
...
     loop: "{{ network['organizations'] }}"
 
+  # Create Storageclass
+  - name: Create Storage Class
+    include_role:
+      name: create/storageclass
+    vars:
+      storageclass_name: "{{ item.cloud_provider }}storageclass"
+      git_dir: "{{ item.gitops.release_dir }}"
+      org: "{{ item }}"
```
sownak, line 35: instead of org = item, use loop_var: org

This applies for all code. All loop should have loop_var so that we don't use default "item"

himanshulalarya: loop_var updated.

Issue #634, opened 2020-04-27
Add channel-artifacts molecule test scenario in Fabric


sownak:
### Description
As a developer I want to run molecule test so that I can test my `create/configtx` role.

### Acceptance Criteria
1.Molecule tests should cover `create/configtx`, `create/channel_artifacts` and related roles
2. Verification step should be configured.
3. Tests should complete successfully

sownak:
This will not be taken up for Sprint 17 due to other tasks taking longer time.

PR #830, opened 2020-07-02
[molecule-test] Add channel-artifacts molecule test scenario in Fabric 

himanshulalarya:
**Changelog**
- Added channel-artifacts molecule test scenario in Fabric

 

**Reviewed by**
@sownak @jagpreetsinghsasan 

 

**Linked issue**
#634


platforms/hyperledger-fabric/configuration/roles/create/configtx/tasks/main.yaml
```diff
@@ -6,16 +6,27 @@
...
+  file:
+    path: "{{ values_dir }}/{{ name }}"
+    recurse: yes
+    state: directory
+
 - name: "Remove old configtx file"
   file:
     path: "{{ config_file }}"
     state: absent
+  changed_when: false
```
sownak, line 22: rather than changed_when false, I believe this should be no_idempotence, as the files are changing, it is just that we always regenerate the config file even when it exists because there is no way to know if anything changed in configuration since the last time.

Issue #967, opened 2020-08-27
Fabric deployment fails 


sownak:
**Describe the bug**
Fabric 1.4.4 and 2.0.0 deployment fails

**To Reproduce**
1. Deploy a Fabric network
2. Deployment fails with following error

The conditional check 'peer.cli == \"enabled\"' failed. The error was: error while evaluating conditional (peer.cli == \"enabled\"): 'dict object' has no attribute 'cli'

**Expected behavior**
Deployment should succeed. Any new parameter added should first check if it is defined.

PR #968, opened 2020-08-27
Cli check updated

himanshulalarya:
Signed-off-by: himanshulalarya <himanshu.a.lal@accenture.com>

**Changelog**
- Add Peer CLI check for backward compatibility so that peer cli can be choice

 

**Reviewed by**
@jagpreetsinghsasan 

 

**Linked issue**
#967

### DIFF
--- a/platforms/hyperledger-fabric/configuration/roles/create/cli_pod/tasks/main.yaml
+++ b/platforms/hyperledger-fabric/configuration/roles/create/cli_pod/tasks/main.yaml
@@ -29,7 +29,7 @@
   loop: "{{ peers }}"
   loop_control:
     loop_var: peer
-  when: peer.cli == "enabled"
+  when: peer.cli is defined and peer.cli == "enabled"
 
 #Git Push : Pushes the above generated files to git directory 
 - name: Git Push
--- a/platforms/hyperledger-fabric/configuration/roles/create/cli_pod/tasks/main.yaml
+++ b/platforms/hyperledger-fabric/configuration/roles/create/cli_pod/tasks/main.yaml
@@ -44,4 +44,4 @@
     GIT_BRANCH: "{{ org.gitops.branch }}"
     GIT_RESET_PATH: "platforms/hyperledger-fabric/configuration"
     msg: "[ci skip] Pushing CLI value files"
- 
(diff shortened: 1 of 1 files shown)


platforms/hyperledger-fabric/configuration/roles/create/cli_pod/tasks/main.yaml
```diff
@@ -29,7 +29,7 @@
   loop: "{{ peers }}"
   loop_control:
     loop_var: peer
-  when: peer.cli == "enabled"
+  when: peer.cli is defined and peer.cli == "enabled"
```
sownak, line 32: add the same condition to git push task as well

jagpreetsinghsasan, line 32: Why do we have 'is defined' check?
The tag will be there in all the peers for sure and it will either be enabled or disabled.

jagpreetsinghsasan (changes requested):

platforms/hyperledger-fabric/configuration/roles/create/cli_pod/tasks/main.yaml
sownak: This is for backward compatibility. cli tag is not mandatory

himanshulalarya: If some network configurations files do not have cli tags, so it will not fail. As peer cli is an additional functionality option

jagpreetsinghsasan: Okay

jagpreetsinghsasan (approved):

sownak (changes requested):
when condition on git push task as well

platforms/hyperledger-fabric/configuration/roles/create/cli_pod/tasks/main.yaml
himanshulalarya: Added

sownak (approved):

Issue #945, opened 2020-08-17
Enable an option to turn off firewall 


jagpreetsinghsasan:
Currently the firewall option to disable it, is not available

The feature which unlinks the float and bridge from deployment based on a condition, and decouples the node from firewall is needed. 
Useful links:
https://docs.corda.net/docs/corda-os/4.5/key-concepts-node.html
https://docs.corda.net/docs/cenm/1.2.html

jwavoetacn:
- Added: `when: peer.firewall.enabled = true` to "float/tasks/main.yaml" and bridge/tasks/main.yaml
- Added the firewall value fields to "/node/values.yaml"

jwavoetacn:
Waiting for #804 which makes it easier to automatically deploy and test the firewall

PR #971, opened 2020-08-27
[ Corda-Ent ] Added option to disable firewall (float & bridge)

jwavoetacn:
**Changelog**
- Added option to disable external firewall float & bridge

**To be reviewed by**
@abevers 

**Linked issue**
#945


sownak (changes requested):
Also remove the setup/bridge and setup/float roles from deploy-network, they should be part of the setup/node role and not directly called from deploy-network because the bridge and float cannot exist without the node.

abevers (commented):
Looks fine to me, when Sownak's comment is implemented, good to merge.

Issue #935, opened 2020-08-17
Investigate the Corda enterprise notary/node log warning 


suvajit-sarkar:
**Describe the bug**
Invalid certificate path warning message on corda enterprise notary/node log 

**To Reproduce**
1. Run BAF corda-ent deployment
2. kubectl get po -A | grep notary to get the notary pod
3. kubectl logs <pod name> -n cenm-ent -c logs
4. See the warning

**Expected behavior**
Investigate the reason for the warning message and fix if possible

**Screenshots**
![image](https://user-images.githubusercontent.com/55580532/90389948-1e8e6f00-e0a8-11ea-9c24-a36343496f71.png)

PR #995, opened 2020-09-04
[fix] corda-enterprise flow and logic

sownak:

**Changelog**
- Remove unused chart pki-node
- Update create/certificates/ambassador to make it more generic so that less code can be used.
- Update deploy-network playbook to make it more relevant for usage

**Next Steps**
- Notary deployment is not working most probably because PVC is not attached for the notary
- Node deployment has not been tested, but needs to follow similar pattern as notary.


sownak:
Notary fixed with following changes:
- Ambassador certs are changed so that unnecessary certificate chaining is not needed. Ambassador tls certs should be independent of the core certificate chain
- idman.crt and networkmap.crt import should be done with correct alias
- sslkeystore should be passed when starting notary as the ketstore is reset when using h2 db 
- networkmap should have idman.crt imported using keytool (hence added the dockerfile to run networkmap as root

Notary is up and running as can be verified by these URLs:
- curl -k https://<nmap public uri>:8443/network-map-user/node-infos
- curl -k https://<nmap public uri>:8443/network-map-user/network-map

platforms/corda-ent/charts/notary/templates/deployment.yaml
```diff
@@ -217,26 +212,40 @@ spec:
...
           chmod +x bin/*
+
+          # to clean network-parameters on every restart
+          rm -rf ${BASE_DIR}/network-parameters
+
+          # add idman and networkmap certificates to java keystore
+          yes | keytool -importcert -file ${BASE_DIR}/certificates/networkmap.crt -storepass changeit -alias {{ .Values.networkServices.networkMapDomain }} -keystore /usr/lib/jvm/zulu-8-amd64/jre/lib/security/cacerts
+          yes | keytool -importcert -file ${BASE_DIR}/certificates/idman.crt -storepass changeit -alias {{ .Values.networkServices.idmanDomain }} -keystore /usr/lib/jvm/zulu-8-amd64/jre/lib/security/cacerts          
           bin/run.sh
+          tail -f /dev/null
```
suvajit-sarkar, line 230: can be removed

platforms/corda-ent/images/networkmap.dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM corda/enterprise-networkmap:1.2-zulu-openjdk8u242
+
+ENV ACCEPT_LICENSE=Y
+USER root
```
suvajit-sarkar, line 4: newline please

suvajit-sarkar:
This PR also resolves the issue raised in #935

jagpreetsinghsasan (approved):

sauveergoel (approved):

platforms/corda-ent/charts/notary/templates/pvc.yaml
```diff
@@ -19,4 +19,4 @@ spec:
   storageClassName: {{ .Values.storage.name }}
   resources:
     requests:
-      storage: 64Mi
+      storage: 512Mi
```
abevers, line 22: Storage size is hardcoded. Move this to `values.yaml` and add to `notary.tpl` file.

platforms/corda-ent/charts/notary/values.yaml
```diff
@@ -85,8 +85,12 @@ rpcSettingsUseSsl: false
 networkServices:
   # Ex. doormanURL: "http://idman-internal.test-kd:10000"
   doormanURL:
+  idmanName:
```
abevers, line 88: Could structure this as such, both are fine
```
  networkServices:
    idman:
      url:
      name: 
      domain:   
    networkmap:
      url:
      name: 
      domain:
```

sownak: These are to be fixed by whoever is doing the notary refactoring

Issue #1003, opened 2020-09-09
Update Peer Address in Cli POD


himanshulalarya:
**Describe the bug**
Peer cli pod links to gossip peer address. Update it to core peer address

**To Reproduce**
1. Run the network
2. Exec into the cli pod
3. Since in single peer gossip address and core peer address are same, it worked but In multi peer setup, it will not execute the commands

**Expected behavior**
Replace gossip peer address with core peer address.

PR #1004, opened 2020-09-09
cli updated

himanshulalarya:
Signed-off-by: himanshulalarya <himanshu.a.lal@accenture.com>

**Changelog**
- Updated peer address in peer cli tpl file.

 

**Reviewed by**
@lakshyakumar 

 

**Linked issue**
#1003


lakshyakumar (changes requested):
seems to be more changes and files that is not mentioned in changelog, other changes are also getting pushed with the required change for the bug.

PR #1007, opened 2020-09-09
[corda-ent] Fix for nodes and flow

sownak:
**Changelog**
- Fix node registration and node charts
- Fix node creation logic
- Fix ambassador certificates creation
- Revert Fabric 2.x channel configurations as they are impacting existing code


platforms/corda-ent/configuration/roles/delete/vault_secrets/tasks/main.yaml
```diff
@@ -109,32 +109,32 @@
...
+    vault delete secret/{{ org.name | lower }}/{{ peer.name | lower }}/certs/{{ network | json_query('orderers[?type==`idman`].name') | first }}
+    vault delete secret/{{ org.name | lower }}/{{ peer.name | lower }}/certs/{{ network | json_query('orderers[?type==`networkmap`].name') | first }}
+    vault delete secret/{{ org.name | lower }}/{{ peer.name | lower }}/certs/networkmaptruststore
+    vault delete secret/{{ org.name | lower }}/{{ peer.name | lower }}/certs/nodekeystore
+    vault delete secret/{{ org.name | lower }}/{{ peer.name | lower }}/certs/sslkeystore
+    vault delete secret/{{ org.name | lower }}/{{ peer.name | lower }}/certs/truststore
+    vault delete secret/{{ org.name | lower }}/{{ peer.name | lower }}/tlscerts
+    vault delete secret/{{ org.name | lower }}/{{ peer.name | lower }}/certs
+    vault delete secret/{{ org.name | lower }}/{{ peer.name | lower }}/root/certs
+    vault delete secret/{{ org.name | lower }}/{{ peer.name | lower }}/network-parameters
```
jagpreetsinghsasan, line 122: This needs to be updated in the docs as well

jagpreetsinghsasan (changes requested):

platforms/corda-ent/charts/node/templates/deployment.yaml
```diff
@@ -209,8 +209,10 @@ spec:
           cp CM-FILES/*.sh bin/
           cp corda.jar bin/
           chmod +x bin/*
-          yes | keytool -importcert -file {{ .Values.workspacePath }}/certificates/tlscerts/networkmap.crt -storepass changeit -alias networkmap -keystore /usr/lib/jvm/zulu8.33.0.1-jdk8.0.192-linux_musl_x64/jre/lib/security/cacerts
-          yes | keytool -importcert -file {{ .Values.workspacePath }}/certificates/tlscerts/idman.crt -storepass changeit -alias idman -keystore /usr/lib/jvm/zulu8.33.0.1-jdk8.0.192-linux_musl_x64/jre/lib/security/cacerts
+          # to clean network-parameters on every restart
```
jagpreetsinghsasan, line 212: Line 147, 152 needs to be changed
IDMAN_CERT=$(echo ${LOOKUP_SECRET_RESPONSE} | jq -r '.data["idman.crt"]')
As here the name of the idman.crt should be like {{ peer.name }}.crt

platforms/corda-ent/configuration/roles/delete/vault_secrets/tasks/main.yaml
sownak: Is part of documentation changes I am working now.

platforms/corda-ent/charts/node/templates/deployment.yaml
sownak: No, why? Why should it be peer.name.crt? These are idman public certs and not peer certs

sauveergoel: Because, in vault the crts are being stored as peer.name.crt, if we name idman and nmap differently, it is not able to pick from vault

sauveergoel: image

platforms/corda-ent/configuration/roles/delete/vault_secrets/tasks/main.yaml
jagpreetsinghsasan: ok

platforms/corda-ent/charts/node/templates/deployment.yaml
jagpreetsinghsasan: Ya, I mean that it should be the actual idman name and networkmap name

sauveergoel (approved):
LGTM

suvajit-sarkar (approved):

platforms/corda-ent/charts/node/templates/deployment.yaml
sownak: Yes, then we need to pass the names in the values.yaml (which I think we are already doing) and use those variables here

PR #1063, opened 2020-09-25
[corda-ent] Fix Helm3 for Corda-enterprise

sownak:
**Changelog**
- Add latest from develop branch
- Fix Flux install and uninstall


platforms/r3-corda-ent/configuration/roles/helm_component/templates/create_signer.tpl
```diff
@@ -0,0 +1,77 @@
+apiVersion: flux.weave.works/v1beta1
```
suvajit-sarkar, line 1: this can be removed.. file has been renamed to signer.tpl

suvajit-sarkar: also is this file required ?
configuration/roles/helm_component/templates/helm_component.tpl

sownak: yes, thats the default template in case no match is found

suvajit-sarkar: ok LGTM then

PR #1159, opened 2020-11-08
[Fabric][fix] Install chaincode with lang node

NthMetal:
Signed-off-by: NthMetal <832529@gmail.com>

**Changelog**
- Fix
platforms\hyperledger-fabric\charts\install_chaincode\templates\install_chaincode.yaml
CC_SRC_PATH is not resolved when lang: "node". This causes install to fail when using node chaincode.
Updated yaml to move chaincode and resolve CC_SRC_PATH

**Reviewed by**
@developer_github_id


jagpreetsinghsasan (changes requested):
Please mention in the network.yaml comments (where this field is mentioned) that peer.chaincode.lang can also be node.
Rest looks fine.

jagpreetsinghsasan (approved):
Looks good to me

Issue #1147, opened 2020-11-03
Add the support for chaincode invocation of init function after the chaincode is committed in fabric version 2.2.x


lakshyakumar:
### Description
As a developer I want to Invoke the `init` function for the chaincode that has been commited to the application channel using the playbook `chaincode-install-instantiate.yaml`.
More description can be found [here](https://hyperledger-fabric.readthedocs.io/en/release-2.2/deploy_chaincode.html#invoking-the-chaincode)

### Steps
1. go through the documentation [here](https://hyperledger-fabric.readthedocs.io/en/release-2.2/deploy_chaincode.html#invoking-the-chaincode)
2. Add the changes to to the Invoke chaincode chart for fetcing of peer certificates and invoking the chaincode for fabric version 2.2
3. Edit the `chaincode-install-instantiate.yaml` playbook to execute the invoke task after the commit chaincode.

### Acceptance Criteria
1. Chaincode is invoked after the chaincode commit using the chaincode-install-instantiate.yaml playbook
2. Test the invocation with fabric version 2.2

NthMetal:
Is this actively being worked on? I can create a PR for this as I need this functionality and already have something similar.

sownak:
@NthMetal Thank you for the interest. Yes, this is already in the current sprint.

lakshyakumar:
@NthMetal Thank you for the interest. This issue is assigned to you, you can start working on it.

PR #1175, opened 2020-11-12
feat/fabric220 initial invoke chaincode (#1147)

NthMetal:
**Changelog**
- Update
Invoke chaincode job added as task after commit chaincode.
Invoke chaincode job is similar to commit chaincode job.
Only runs on type == 'creator'.
Checks to make sure commit chaincode job was successful.
Checks if invoke-chaincode is already run
Writes all endorsing orgs certificates to vault
Creates value file for chaincode invocation
Pushes generated value file to git directory 

**Reviewed by**
@lakshyakumar

 

**Linked issue**
#1147


NthMetal:
Closed because something went terribly wrong after I forgot to commit with "-s".

Issue #1166, opened 2020-11-10
FABRIC-install chaincode job fails due incorrect chaincode path


bafsunsu:
**Describe the bug**

Chaincode installation jobs fails for fabric 2.0.0 with error as 
"Error: error getting chaincode deployment spec for CHAINCODE_NAME"
For "platforms/hyperledger-fabric/charts/install_chaincode/templates/install_chaincode.yaml" at line number 154 we perform a change directory with "cd $GOPATH/src/github.com/chaincode && dep ensure"  command but then again on line no 156 for peer chaincode install command we are using the chaincode path variable value as. "-p github.com/chaincode/${CHAINCODE_NAME}/${CHAINCODE_MAINDIR}" which is not reachable from the present directory which is already at $GOPATH/src/github.com/chaincode/ and the install command is not able to find the chaincode to install

Branch:Master

**To Reproduce**
1. Set up a new fabric network with details of chaincode to be installed
2.  At chaincode installation step look for logs of Job "installchaincode-{{ $.Values.peer.name }}-{{ $.Values.chaincode.name }}-{{ $.Values.chaincode.version }}"
3. the job would fail with "Error: error getting chaincode deployment spec for <chaincode_name>: exit status 1
4. Re executing same commands of install_chaincode.yaml file through a peer cli would present the same error

**Expected behavior**
Chaincode to get installed and instantiated on the network

**Screenshots**

**Environment (please complete the following information):**
 - Azure Kubernetes service
 - Version 2.0.0

**Additional context**

lakshyakumar:
hi @bafsunsu , can you please provide us more details on the issue as we are not able to reproduce the same from the master branch.
Also a seperate point being we are not actively supporting Hyperledger-Fabric v2.0.0, currently we are supporting HL Fabric v1.4.4 and v2.2.x is in active development.

bafsunsu:
@lakshyakumar Sorry there was little mistake in the issue description as I had mentioned instantiate_chaincode.yaml files reference instead of install_chaincode.yaml file at "platforms/hyperledger-fabric/charts/install_chaincode/templates/install_chaincode.yaml"

Have updated the issue description and also have created a pull request for this issue fix from my public git hub id(accidentally created issue through baf test github ID)

suvajit-sarkar:
Can you verify is this is a valid bug on Fabric 2.2 latest code

suvajit-sarkar:
[...]

PR #1201, opened 2020-11-29
[ Fabric ]: Fix for chaincode install job

Oreo-Tech:
**Changelog**
- Fix

Description: On line number 158 we are doing change directory to "$GOPATH/src/github.com/chaincode" but on the next command (line no: 160) we were expecting the current directory to be relative to "github.com/chaincode/${CHAINCODE_NAME}/${CHAINCODE_MAINDIR}" path to install the chaincode which is not reachable from the present directory as we already moved at $GOPATH/src/github.com/chaincode/ and the install command is not able to find the chaincode to install
 

**Reviewed by**
@jagpreetsinghsasan 

 

**Linked issue**
#1166

### DIFF
--- a/platforms/hyperledger-fabric/charts/install_chaincode/templates/install_chaincode.yaml
+++ b/platforms/hyperledger-fabric/charts/install_chaincode/templates/install_chaincode.yaml
@@ -157,7 +157,7 @@ spec:
           mkdir -p $GOPATH/bin && curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
           cd $GOPATH/src/github.com/chaincode && dep ensure
           ## Installing chaincode on a peer
-          peer chaincode install -n ${CHAINCODE_NAME} -v ${CHAINCODE_VERSION} -p github.com/chaincode/${CHAINCODE_NAME}/${CHAINCODE_MAINDIR}
+          peer chaincode install -n ${CHAINCODE_NAME} -v ${CHAINCODE_VERSION} -p $GOPATH/src/github.com/chaincode
         env:
         - name: CORE_VM_ENDPOINT
           value: unix:///host/var/run/docker.sock
--- a/platforms/hyperledger-fabric/configuration/roles/create/chaincode/install/tasks/valuefile.yaml
+++ b/platforms/hyperledger-fabric/configuration/roles/create/chaincode/install/tasks/valuefile.yaml
@@ -45,7 +45,7 @@
     type: "install_chaincode_job"
     peer_name: "{{ peer.name }}"
     peer_address: "{{ peer.name }}.{{ namespace }}:{{ peer.grpc.port }}"
-    component_name: "chaincode-install-{{ name }}-{{ peer.name }}-{{ peer.chaincode.version | replace('.','-')}}"
+    component_name: "chaincode-install-{{ name | lower }}-{{ peer.name }}-{{peer.chaincode.name}}{{peer.chaincode.version}}"
     component_chaincode: "{{ peer.chaincode }}"
     fabrictools_image: "hyperledger/fabric-tools:{{ network.version }}"
     alpine_image: "{{ docker_url }}/alpine-utils:1.0"


sownak:
Thanks for the PR. The sign-off is missing. Also has this been tested? Because I think without the main file, the install will not be successful.

sownak (changes requested):
Signoff needs to be added.

jagpreetsinghsasan:
I don't think this will work as the chaincode folder doesn't necessarily have the main file to run.
(For example, fabcar java cc)

Oreo-Tech:
The Fix was majorly in focusing on providing the correct relative path where chaincode is stored "$GOPATH/src/github.com/chaincode" instead of the earlier coded path "github.com/chaincode.." which was not reachable as we are already present in the folder due to "cd $GOPATH/src/github.com/chaincode" instruction on line 158

talking about the removal of "${CHAINCODE_NAME}/${CHAINCODE_MAINDIR}" section from the code, I agree its missing but do we really need mentioning a Main directory which does not play any major role as we already have "Values.chaincode.repository.path " doing the work to provide path for main file to us

Line - 150: cp -R /tmp/chaincode/{{ $.Values.chaincode.repository.path }}/* $GOPATH/src/github.com/chaincode/

all we do is track the chaincode repository path and copy it into local path so I thought to avoid the added dependency of main dir and provide the correct repository path

sownak:
Yes we do need mentioning a Main directory because without that, the chaincode does not install. In your case, the main directory may be the same as the chaincode directory, but that may not be true for everyone else.

platforms/hyperledger-fabric/charts/install_chaincode/templates/install_chaincode.yaml
```diff
@@ -157,7 +157,7 @@ spec:
           mkdir -p $GOPATH/bin && curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
           cd $GOPATH/src/github.com/chaincode && dep ensure
           ## Installing chaincode on a peer
-          peer chaincode install -n ${CHAINCODE_NAME} -v ${CHAINCODE_VERSION} -p github.com/chaincode/${CHAINCODE_NAME}/${CHAINCODE_MAINDIR}
+          peer chaincode install -n ${CHAINCODE_NAME} -v ${CHAINCODE_VERSION} -p $GOPATH/src/github.com/chaincode
```
sownak, line 160: What if there are multiple chaincodes in the $GOPATH/src/github.com/chaincode/ directory? How will the install chaincode work then? And each chaincode can have a different main directory, as I mentioned in the comment.

sownak (changes requested):

platforms/hyperledger-fabric/charts/install_chaincode/templates/install_chaincode.yaml
bafsunsu: thanks, got your point, am making the changes accordingly and will update it

platforms/hyperledger-fabric/configuration/roles/create/chaincode/install/tasks/valuefile.yaml
```diff
@@ -45,7 +45,7 @@
     type: "install_chaincode_job"
     peer_name: "{{ peer.name }}"
     peer_address: "{{ peer.name }}.{{ namespace }}:{{ peer.grpc.port }}"
-    component_name: "chaincode-install-{{ name }}-{{ peer.name }}-{{ peer.chaincode.version | replace('.','-')}}"
+    component_name: "chaincode-install-{{ name | lower }}-{{ peer.name }}-{{peer.chaincode.name}}{{peer.chaincode.version}}"
```
arsulegai, line 48: Please add back `| replace('.','-')` to the version.

arsulegai (changes requested):

suvajit-sarkar:
closing this PR as it has gone stale, if required create a new PR to address the issue

Issue #1150, opened 2020-11-03
Setup Float on a seperate cluster


jagpreetsinghsasan:
### Description
As a developer, I want to install float on a separate cluster.

1. Dependent on #1149 
2. Update the network.yaml with the new changes
![image](https://user-images.githubusercontent.com/56861721/97978543-b6703f00-1df3-11eb-86b4-3ef4fafed2ff.png)
![image](https://user-images.githubusercontent.com/56861721/97978611-cee05980-1df3-11eb-973c-5c69533f8ab2.png)
3. Separate out the float role from the node role
4. The pre-setup done for node (namespace, service accounts, rbac, vault configuration, ambassador, flux) should happen for float as well (exactly the same thing, but for the new cluster)
5. After the pki job completes for node, the firewallca.jks, float.jks and bridge.jks should be placed in the float vault as well via ansible controller
6. Finally, the float deployment should happen on the new cluster (make sure the release directory should not be in sync with the flux of the node cluster, otherwise  both the node cluster and the float cluster will have floats setup)
7. Reset for external float shall also happen

Note: Make sure, the ambassador on the float cluster has the ports open (2 required per float) and the float url has the external URL of the float cluster (even in the node-registration, node and bridge tpl files)

### Acceptance Criteria
Float should be up and running in a different cluster, if the float cloud credentials are different from that of the node cloud credentials

PR #1202, opened 2020-11-30
[ Corda-Ent ] Enable Float in serparate cluster and vault for corda enterprise

suvajit-sarkar:
Signed-off-by: suvajit-sarkar <suvajit.sarkar@accenture.com>

**Changelog**
- Add firewall (bridge external) and float in separate cluster and vault option
  - Removed the TLS context for float service which was preventing ssl pass through
  - Update the node pki gen chart to put firewall certs to float vault as well
  - Update node registration chart to put node ssl and trust store to float vault
  - Updated the ansible roles for firewall enabled tasks
- Update the Corda configuration doc and sample file 
- Update the reset network to cleanup both network and float cluster and vaults

 

**Reviewed by**
@jagpreetsinghsasan @abevers 

 

**Linked issue**
#1150


jagpreetsinghsasan (approved):
LGTM

sownak:
How are we testing this via the "baf-configurations"?

suvajit-sarkar:
@sownak we have not test it via Jenkins / BAF configuration yet. We need to update the Jenkins script to provide additional vault address and kubeconfig for the float to test it in separate cluster and vault (might also have to look at cluster access). Though the code will work perfectly fine if we keep float cluster and vault same as that of the organizations. I will add the updated network yaml to the baf-configurations and test this out.

PR #1240, opened 2020-12-15
Jenkins script update 

suvajit-sarkar:
Signed-off-by: suvajit-sarkar <suvajit.sarkar@accenture.com>

**Changelog**
- Added "removal of an org" in Jenkins stage for Hyperledger Fabric for v2.2.0
- Replaced version 2.0.0 with with 2.2.0 as Fabric version choice from Jenkins build options

 

**Reviewed by**
@jagpreetsinghsasan


jagpreetsinghsasan (approved):
LGTM

automation/Jenkinsfile
```diff
@@ -375,7 +379,37 @@ pipeline {
...
     }
-    
+    /*
+    *  Execute remove-organization.yamll which will remove an organisation from the network
+    */
+    stage('Remove an organisation') {
+      when {
+        not {
+          anyOf {
+            branch 'master';
```
sownak, line 389: instead do when fabric?

sownak (changes requested):

automation/Jenkinsfile
suvajit-sarkar: yep will do, was thinking of gradually removing the not branches when we do for other DLTs , but yeah for now just Fabric makes more sense

Issue #1160, opened 2020-11-09
Upgrade Besu to latest stable


suvajit-sarkar:
### Description
As a developer I want to upgrade besu to latest version v.1.5.x so that we have the latest feature of DNS support in Besu.

#### Link for reference
https://github.com/hyperledger/besu/releases/tag/1.5.3

### Acceptance Criteria
1.  Network up and running with Besu v1.5.3
2.  Deploy and test the supplychain refapp

PR #1248, opened 2020-12-18
[HL Besu] Upgraded HL Besu to v1.5.5 and Orion to v1.6.0

sauveergoel:
Signed-off-by: sauveergoel <sauveer.goel@accenture.com>

**Changelog**
- Upgraded HL Besu to v1.5.5 and Orion to v1.6.0 

**Reviewed by**
@sownak @suvajit-sarkar @lakshyakumar  

**Linked issue**
#1160

### DIFF
--- a/docs/source/operations/besu_networkyaml.md
+++ b/docs/source/operations/besu_networkyaml.md
@@ -28,7 +28,7 @@ The sections in the sample configuration file are
 
 `type` defines the platform choice like corda/fabric/indy/quorum/besu, here in the example its **besu**.
 
-`version` defines the version of platform being used. The current Hyperledger Besu version support is only for **1.4.4**.
+`version` defines the version of platform being used. The current Hyperledger Besu version support is only for **1.5.5**.
 
 
 `env` section contains the environment type and additional (other than 8443) Ambassador port configuration. Vaule for proxy field under this section can be 'ambassador' as 'haproxy' has not been implemented for Besu.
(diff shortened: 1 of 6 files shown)


lakshyakumar (approved):
LGTM

docs/source/operations/besu_networkyaml.md
```diff
@@ -28,7 +28,7 @@ The sections in the sample configuration file are
 
 `type` defines the platform choice like corda/fabric/indy/quorum/besu, here in the example its **besu**.
 
-`version` defines the version of platform being used. The current Hyperledger Besu version support is only for **1.4.4**.
+`version` defines the version of platform being used. The current Hyperledger Besu version support is only for **1.5.5**.
```
sownak, line 31: are we removing support for 1.4.4?

platforms/hyperledger-besu/charts/node_validator/templates/deployment.yaml
```diff
@@ -184,8 +185,6 @@ spec:
             containerPort: {{ .Values.node.ports.p2p }}
             protocol: TCP
         env:
-          - name: BESU_GENESIS_DIR
```
sownak, line 187: this will delete the genesis file when the pod is restarted. Is that what we want?

docs/source/operations/besu_networkyaml.md
sauveergoel: yes, since the besu team is not actively developing multiple versions and the network is forward compatible, we are removing the 1.4.4 support. 
10e6ac61-236d-4799-8dd8-0a78600b1e6e

platforms/hyperledger-besu/charts/node_validator/templates/deployment.yaml
sauveergoel: In the docker image for v1.5.5, a user named besu has been introduced, which is not having sudo permissions and can only have rights to write in the /tmp folder. the value of the BESU_GENESIS_DIR is being picked up from mountPath which is being used at other places too. hence didn't want to touch it or edit it, which is the reason why hardcoded the genesis dir path.
The genesis file will not get deleted as it is being picked up from the vault when the pod starts.

sownak (approved):

Issue #1294, opened 2021-02-05
Organization that is not part of the channel is not added to orderer genesis


rahultoraskarr:
**Describe the bug**
I am trying to create Hyperledger Fabric network with 3 orgs, say A,B and C. I need two channels on the network, one with orgs (A, B) and other with orgs (A, C).

Due to limitation in BAF, I can only specify one channel when creating a network. When creating network with all 3 orgs and only A and B in channel, org C doesn't get added to consortium or genesis block. Due to which when running playbook 'add-new-channel.yaml' to create second channel (A, C) I get below error

Error: got unexpected status: BAD_REQUEST -- Attempted to include member C which is not in the consortium

**To Reproduce**
Create a Fabric network with 3 orgs and add only two orgs to the channel

**Expected behavior**
All orgs in the network should be part of consortium 

**Environment (please complete the following information):**
 - Version : 2.2.0

alvaropicazo:
![image](https://user-images.githubusercontent.com/76157062/107761043-03d14600-6d2b-11eb-9960-04c91b6ee4f7.png)

This is how the system channel looks when there are three Orgs in the network and two Orgs in the channel.

alvaropicazo:
A new issue has been created related to adding a new channel to the existing network, as also was part of the error that was being faced in this issue. To see more details regarding the issue and have a tracking, can be done here: #1304

PR #1302, opened 2021-02-12
[ Fabric ] All Organizations are now part of the Consortium  #1294

alvaropicazo:
Signed-off-by: alvaropicazo <alvaro.picazo.haase@accenture.com>

**Changelog**
- Updated configtxProfile.tpl so now it takes every existing organization, previously it was taking the orgs for the channel that was created, so the ones which were not assigned to that channel were being excluded.
- Fixed git_ssh parameter being replaced by git_url (now in the network.yaml files, in the gitops section, we are using git_url instead of git_ssh).
- Updated files for TLS CAs which where failing when approving the chaincode, commiting, upgrading the chaincode for fabric v2.2.

 

**Reviewed by**
@jagpreetsinghsasan 
@suvajit-sarkar 

 

**Linked issue**
#1294


suvajit-sarkar (approved):

jagpreetsinghsasan (approved):

platforms/hyperledger-fabric/configuration/roles/create/configtx/templates/configtxProfile.tpl
```diff
@@ -20,8 +20,10 @@ Profiles:
     Consortiums:
       {{channel.consortium}}:
         Organizations:
-{% for org in channel.participants %}
+{% for org in network.organizations %}
```
arsulegai, line 23: Ideally this should be documented. This would allow one to add consortium members just by keeping those names in the `network.yaml` while creating the channel.

I was thinking about providing this option as an edit to config when required instead.

Issue #1316, opened 2021-03-01
Introduce admin and orderer Node OUs


arsulegai:
**Is your feature request related to a problem? Please describe.**
Make use of orderer and admin Node OUs. This allows any new certificate that has `admin` role being able to use for administration purpose, need not be stored in the admincerts directory of the MSP. Allowing multiple certs to be generated on demand for a given MSP.

**Describe the solution you'd like**
Add support of `orderer` and `admin` Node OUs in the config file.

**Describe alternatives you've considered**
None

**Additional context**
`orderer` and `admin` Node OUs were not supported at the time of writing BAF for Fabric.

sownak:
This issue will be solved by PR #1393

PR #1393, opened 2021-04-09
[fabric] Add peer over multiple clusters

### LINKED ISSUE #1318: Fabric peer nodes across clusters
**Is your feature request related to a problem? Please describe.**
Hyperledger Fabric supports multiple peers belonging to one organization to be on the channel. The feature request is to add more than one peer of an organization MSP, but they are installed to different Kubernetes clusters.

**Describe the solution you'd like**
There can be many ways to achieve this, but here is what I think can be done faster
1. Create/join a channel from one cluster
2. Create a new peer on another cluster and join the channel
3. Handle Vault/CA dependency

**Describe alternatives you've considered**
None

**Additional context**
None

sownak:

**Changelog**
- Add support for Adding new Peer on a separate Kubernetes cluster
- Fix adding peer on existing cluster as well
- Fix new id.type when generating certificates 

**Reviewed by**
@alvaropicazo  

**Linked issue**
Resolves #1318 and #1316


docs/source/operations/adding_new_peer_fabric.md
```diff
@@ -64,18 +66,23 @@ and under `network.organizations` as
...
-
+The `network.yaml` file should contain the specific `network.organization` patch. Orderer information is needed if you are going to install/upgrade the existing chaincodes, otherwise it is not needed. And the `org_status` must be `existing` when adding peer.
 
-For reference, see `network-fabric-add-peer.yaml` file [here](https://github.com/hyperledger-labs/blockchain-automation-framework/tree/master/platforms/hyperledger-fabric/configuration/samples).
+Ensure the following is considered when adding the new peer on a different cluster:
+- The CA server is accessible publicly or at least from the new cluster.
+- The CA server public certificate is stored in a local path and that path provided in network.yaml.
+- There is a single Hashicorp Vault and both clusters (as well as ansible controller) can access it.
+- Admin User certs have been already generated and store in Vault (this is taken care of by deploy-network.yaml playbook if you are using BAF to setup the network).
+- The `network.env.type` is different for different clusters.
```
arsulegai, line 76: Is this necessary?
`env.type` can be used to identify `dev`, `stage`, `prod`. Having a different type for the same network across clusters would not be preferred. Is this done because of flux?

Would it help if flux is tied with the github?

sownak: Yes, this is tied to flux, if using the same git-repo branch for different clusters, the `type` must be different otherwise flux sync will get confused because flux tags the branch with a sync tag which has to be unique for different flux.

sownak: If prod is over multiple clusters, the tag can be called prod1, prod2 etc

arsulegai: We can make the type as namespace for the flux and associate the flux sync with the github section (either the release folder or the github link etc.). This can be taken up as part of running flux in its own namespace.

Issue #723, opened 2020-06-04
Add crypto-constellation molecule test scenario in Quorum


sownak:
### Description
As a developer I want to run molecule test so that I can test my `create/crypto/constellation` role.

### Acceptance Criteria
1.Molecule tests should cover `create/crypto/constellation` and related roles
2. Verification step should be configured.
3. Tests should complete successfully

Dimbu-Afonso:
Looking into this

PR #1469, opened 2021-05-19
Added molecule test for crypto-constellation and added ignore-yes for…

Dimbu-Afonso:
… vault related task since we're not testing vault, also added idempotence tags

Signed-off-by: Dimbu Afonso <dimbu.k.afonso@accenture.com>

**Changelog**
- Added molecule test for crypto-constellation and added ignore-yes for vault related tasks since we're not testing vault - Moreover, added idempotence tags.

 

**Reviewed by**
@suvajit-sarkar @sownak 

 

**Linked issue**
#723

### DIFF
--- a/platforms/quorum/configuration/roles/create/crypto/constellation/tasks/nested_main.yaml
+++ b/platforms/quorum/configuration/roles/create/crypto/constellation/tasks/nested_main.yaml
@@ -7,6 +7,8 @@
     VAULT_TOKEN: "{{ vault.root_token }}"
   register: vault_tm_result
   ignore_errors: yes
+  tags:
+    molecule-idempotence-notest
 
 # This task creates the build directory if it does not exist
 - name: "Creating build directory"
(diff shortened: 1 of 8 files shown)


jagpreetsinghsasan:
A similar comment here,
Why are we having no idempotence checks in all tasks except the last one?

platforms/quorum/configuration/roles/create/crypto/constellation/tasks/nested_main.yaml
```diff
@@ -30,3 +36,4 @@
     VAULT_ADDR: "{{ vault.url }}"
     VAULT_TOKEN: "{{ vault.root_token }}"
   when: vault_tm_result.failed == True
+  ignore_errors: yes
```
sownak, line 39: Why is ignore error added here? This should fail if the vault cannot be updated

sownak (changes requested):

Issue #1462, opened 2021-05-13
Create ambassador tls secret in organization namespace


suvajit-sarkar:
### Description
As a developer I want to create the ambassador tls secret in the organization namespace instead of default namespace for corda enterprise and OS

(add more description and relevant links)

### Acceptance Criteria
1. Network deploys successfully without error
2. Ambassador tls secrets are create in org namespace
3. The services connect over tls

pppazos:
I can work this

Issue #996, opened 2020-09-07
Integrate Ambassador proxy certificates with PKI-tool certificates for Corda Enterprise


sownak:
### Description
As a developer I want to generate/integrate the Ambassador proxy certificates using PKI-tool so that the Ambassador proxy uses the certificates generated from the same certification authority as the Corda Nodes.

1. Explore how pki generator can be used to create Ambassador certificates with the public domain names like idman.xx.domain.com etc.
2. This will ensure the ambassador certificates are in the truststore already, hence we do not need to add them via `keytool -importcert -file ${BASE_DIR}/certificates/networkmap.crt ...` commands
3. `platforms/corda-ent/configuration/roles/create/certificates/cenm/tasks/main.yaml` need to be modified to extract the cert and key from keystore and update the kubernetes secret for ambassador
4. Then we do not need root access for the docker images of networkmap, notary and node.

### Acceptance Criteria
1. Ambassador certificates are generated using pki-generator and not openssl
2. No "keytool -importcert -file ${BASE_DIR}/certificates/networkmap.crt .." commands in charts
3. Notary and Networkmap is updated without any error.
4. Notary and Nodes are visible when the following is called
- curl -k https://nmap-public-uri:8443/network-map-user/node-infos
- curl -k https://nmap-public-uri:8443/network-map-user/network-map

sownak:
@jagpreetsinghsasan  this is the issue to remove openssl certs from Corda-enterprise

pppazos:
working on this

PR #1501, opened 2021-06-03
[corda-ent] certificates for ambassador

pppazos:
**Changelog**
- in cenm (idman, network, notaries) trust tls root ca in cenm organization
- in nodes (include idman and network crts in corda truststore)
- k8s tls secrets created in the org namespace

**Reviewed by**
@sownak 
@suvajit-sarkar 

 

**Linked issue**
#996 #1462

### DIFF
--- a/platforms/r3-corda-ent/charts/node/templates/deployment.yaml
+++ b/platforms/r3-corda-ent/charts/node/templates/deployment.yaml
@@ -393,7 +393,6 @@ spec:
               echo
               sleep ${HOW_LONG}
           fi
-          sleep 1200
           echo "DONE"
         volumeMounts:
         - name: node-volume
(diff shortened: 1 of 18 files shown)


platforms/r3-corda-ent/charts/node/templates/deployment.yaml
```diff
@@ -388,6 +393,7 @@ spec:
               echo
               sleep ${HOW_LONG}
           fi
+          sleep 1200
```
sownak, line 396: This can be removed

sownak (changes requested):
rest looks fine, has the whole application been tested after these changes?

platforms/r3-corda-ent/charts/node/templates/deployment.yaml
pppazos: * Sleep deleted.
* For the testing, a corda-ent network was deployed (cenm v1.5) with 3 node orgs ( no firewall and no cordapp). Validated using the endpoint /network-map-user/node-infos that the 2 notaries and 3 nodes were registered.

sownak (approved):

PR #1595, opened 2021-08-03
[quorum] Update to vault v2

kaverikhaneja:
Signed-off-by: kaverikhaneja <kaveri.khaneja@accenture.com>

**Changelog**

- Update vault commands to v2
- Update any file paths, as per needed

 

**Reviewed by**
@suvajit-sarkar 
@sownak 

 

**Linked issue**
#1549

### DIFF
--- a/platforms/hyperledger-besu/configuration/roles/create/crypto/tessera/Readme.md
+++ b/platforms/hyperledger-besu/configuration/roles/create/crypto/tessera/Readme.md
@@ -1,76 +1,69 @@
 ## ROLE: create/crypto/tessera
 This role generates the crypto material for tessera transaction manager
 
-### Tasks
-(Variables with * are fetched from the playbook which is calling this role)
-#### 1. Call nested_main for each node in an organisation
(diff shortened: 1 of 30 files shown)


platforms/quorum/configuration/roles/create/crypto/tessera/tasks/nested_main.yaml
```diff
@@ -0,0 +1,60 @@
+# This task checks if the tm crypto material is already stored in the vault or not
```
sownak, line 1: Dont think this file is needed now

sownak (changes requested):

platforms/quorum/configuration/roles/create/crypto/tessera/tasks/nested_main.yaml
kaverikhaneja: You're right. I will remove it and update the Readme.md as well.

sownak (approved):

PR #1709, opened 2021-10-12
[fabric] Fix CA server and user certificate issues

sownak:

**Changelog**
- Fix CA server not starting
- Rename playbook to include verb
- Update logic to store user certs to Vault

### DIFF
--- a/platforms/hyperledger-fabric/configuration/roles/create/crypto/peer/tasks/main.yaml
+++ b/platforms/hyperledger-fabric/configuration/roles/create/crypto/peer/tasks/main.yaml
@@ -105,6 +105,7 @@
     VAULT_TOKEN: "{{ vault.root_token }}"
   register: vault_admin_result_yaml
   when: not vault_admin_result.failed
+  changed_when: false
 
 # common.yaml should be called only once, do not add a loop on peers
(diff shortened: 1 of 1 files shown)


platforms/hyperledger-fabric/configuration/roles/create/users/tasks/save_to_vault.yaml
```diff
@@ -0,0 +1,37 @@
+# Check user msp already created
+- name: Check if user msp already created
+  shell: |
+    vault kv get -field=admincerts {{ vault.secret_path | default('secretsv2') }}/crypto/peerOrganizations/{{ component_name }}/users/{{ user.identity }}/msp
+  environment:
+    VAULT_ADDR: "{{ vault.url }}"
+    VAULT_TOKEN: "{{ vault.root_token }}"
+  register: vault_user_result
```
arsulegai, line 8: Ideally, we should also copy newly generated certificates (for the same key) in case of a refresh. The generate user crypto script has been modified to refresh an enrolment as needed. For example, a certificate gets refreshed if it is about to expire in next X hours, a certificate gets refreshed if for a given user the attributes change.

Issue #1761, opened 2021-11-22
Ensure successful storage of cryto materials in vault by Fabric helm charts


suvajit-sarkar:
### Description
As a developer I want to check and ensure that the store certs are successfully storing the cryto materials in vault.

### Acceptance Criteria
1. Checks to ensure cryto successfully stored
2. Retry logic based on the above check
3. Add retry logic for reading from vault

TonyRowntree:
I am taking this

mgCepeda:
i am working in this issue

PR #1812, opened 2022-01-17
[fabric] Ensure successful storage of crypto materials in vault

mgCepeda:
Signed-off-by: mgCepeda <marina.gomez.cepeda@accenture.com>

**Changelog**
- add in ca-tools chart logic to check the storage of cryptographic materials in the vault
- add a new role check/crypto_materials

 

**Reviewed by**
@suvajit-sarkar
@jagpreetsinghsasan

 

**Linked issue**
#1761

### DIFF
--- a/platforms/hyperledger-fabric/configuration/roles/create/ca-tools/orderer/tasks/main.yaml
+++ b/platforms/hyperledger-fabric/configuration/roles/create/ca-tools/orderer/tasks/main.yaml
@@ -8,7 +8,7 @@
 # This role creates value file for the deployment of CA Tools CLI
 #############################################################################################
 
-# Check if CA server is available
+# # Check if CA server is available
 - name: "waiting for the CA server to be created in {{ item.name | lower }}-net"
   include_role:
     name: "{{ playbook_dir }}/../../shared/configuration/roles/check/helm_component"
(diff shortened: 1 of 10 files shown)


platforms/hyperledger-fabric/configuration/roles/create/ca-tools/orderer/tasks/main.yaml
```diff
@@ -68,6 +68,16 @@
   tags:
     - notest
 
+#####################################################################################################################
+# This task Check if crypto materials exists in vault.
+- name: "Check if crypto materials exists in vault."
+  include_role:
+    name: check/crypto_materials
+  vars: 
+    namespace: "{{ item.name | lower }}-net"
```
sownak, line 77: What is the point of this check from Ansible? It is not telling the user what is missing and will succeed anyway. The Issue should only check in the crypto creation job, and fail the job if the secrets were not saved.

jagpreetsinghsasan: This role contains waits for various certs like orderertls, msp etc
So, what I have asked @mgCepeda to do is, to have this task in `shared/configuration/roles/check/setup`
image

And then flatten out the contents of `check/crypto_materials` into this same file `platforms/hyperledger-fabrc/configuration/roles/ca-tools/orderer/tasks/main.yaml` as these
image

sownak: So has the wait for crypto jobs completion been removed?

jagpreetsinghsasan: We need the waits, because the crypto creation happens via one of the containers of `ca-tools`

sownak (approved):

suvajit-sarkar (approved):

PR #1939, opened 2022-05-09
[fabric] Multiple operations fix after ansible decoupling changes

prasanths96:
**Changelog**
- **Add peer to org:** 
    - `platforms/hyperledger-fabric/configuration/add-peer.yaml`: "create/crypto/peer" role was no more available & replaced
    - `platforms/hyperledger-fabric/charts/catools/templates/configmap.yaml`: store-vault-peer.sh logic fix: If peer status is not "new" or "", then move on to next peer.
    - `platforms/hyperledger-fabric/configuration/roles/create/ca-tools/peer/tasks/main.yaml`:  The export syntax: 
    `export VARIABLE=$VALUE` does not seem to work in the sh used in bevel. Instead two line approach works `VARIABLE=$VALUE; export VARIABLE`. Ref: [Link](https://stackoverflow.com/questions/7328223/unix-export-command)
    
- **Add raft orderer to org:**
    - `platforms/hyperledger-fabric/configuration/add-orderer.yaml`: "create/crypto/orderer" role was no more available & replaced
    - `platforms/hyperledger-fabric/configuration/roles/create/ca-tools/orderer/tasks/main.yaml`: Export syntax change
    
- **Removing org:**
    - `platforms/hyperledger-fabric/configuration/roles/setup/config_block/sign_and_update/tasks/main.yaml`: Changed condition to also involve the org being deleted to sign the config update block. 
    In a 3 org setup `{supplychain (orderer), carrier, manufacturer}`, removing one of the peer orgs lead to error indicating channel update policy not being met. Making the above change fixes this.

### DIFF
--- a/platforms/hyperledger-fabric/configuration/add-orderer.yaml
+++ b/platforms/hyperledger-fabric/configuration/add-orderer.yaml
@@ -33,21 +33,21 @@
       orderers: "{{ item.services.orderers }}"
     loop: "{{ network['organizations'] }}"
     when: item.type == 'orderer'
-    
-  # Create Orderer crypto materials
-  - name: "Generate crypto for new orderer"
+  
+  # Create CA Tools helm-value files and check-in
+  - name: Create CA tools for each organization
     include_role:
-      name: "create/crypto/orderer"
+      name: "create/ca-tools/orderer"
     vars:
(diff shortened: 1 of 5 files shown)


prasanths96:
Updating fork with new commits before PR

PR #1959, opened 2022-05-25
[fabric] Fix for helm operator to fetch chart from private git repo using https

pras-spydra:
This resolves the issue of bevel network failing with private git repositories. 

**Reason for the failures:** 
- flux-helm-operator wasn't configured used/configured correctly with git credentials. So, it was not able to fetch the charts from the private git repo as defined under the `chart:`  section of the value file. 

**Fix:**
- Create a kubernetes secret which has the git credentials stored inside the variables: `username` and `password` (within each of the organization namespaces respectively)
- Add `secretRef:` field in the value files and refer the above mentioned secret.

**Doc Reference:** https://fluxcd.io/legacy/helm-operator/helmrelease-guide/chart-sources/#https

Signed-off-by: Prasanth Sundaravelu <prasanth.s@spydra.app>


platforms/shared/configuration/roles/check/helm_component/tasks/main.yaml
```diff
@@ -52,4 +52,4 @@
   retries: "{{ network.env.retry_count}}"
   delay: "{{ delay | default(30) }}"
   until: component_data.resources|length > 0
-  when: component_type == "Pod"
```
jagpreetsinghsasan, line 55: Just a new line here.

platforms/shared/configuration/roles/setup/flux/tasks/main.yaml
```diff
@@ -84,4 +84,4 @@
     label_selectors:
       - app = flux
       - release = flux-{{ network.env.type }}
-  when: flux_service.resources|length == 0
```
jagpreetsinghsasan, line 87: Just a new line here

jagpreetsinghsasan (changes requested):
Just a comment, at some places it says `git_protocol` and at some places it says `git_ptc`. Can we have it the same everywhere?

pras-spydra:
[jagpreetsinghsasan](https://github.com/jagpreetsinghsasan) In the places where it is "git_ptc", it is being passed off into another role, which re-assigned the same variable to "git_protocol" and in turn actually passes to the template. 
Template is the final consumer of the variable "git_protocol". When I used the same name "git_protocol" with the re-assignment, ansible threw "recursive assignment" error. 
I am new to ansible, if there is a neater way to fix this, please suggest me.

sownak:
@pras-spydra As we are using all the `gitops.xx` in the tpl files, why not send the whole `gitops` to helm_template role and used `gitops.git_protocol` in the tpl files.

pras-spydra:
@jagpreetsinghsasan I've made the changes. 
@sownak I've changed the `git_ptc` variables to `git_protocol` for now, since passing full `gitops` to template needs changes in a lot of files. I could do it later with a separate PR.

jagpreetsinghsasan:
The changes look fine to me @pras-spydra .
Just one last thing is the molecule tests needs to be passed. I can see the fabric one failing (generally happens when you introduce new variables to the code or edit the code, let me know if you have issues fixing the molecule test cases, I shall help)

pras-spydra:
@jagpreetsinghsasan I probably wouldn't be able to work on this for this week, as my schedule is tight. I will try to do as soon as possible. Sorry for delaying.

jagpreetsinghsasan (approved):

jagpreetsinghsasan (changes requested):
@pras-spydra thanks for the PR. Its fine, the molecule tests have been fixed in the develop branch as @suvajit-sarkar mentioned.Now if you rebase your commit over the latest develop branch code, the PR will have a single commit with all your contribution in it with regards to this PR.

jagpreetsinghsasan (approved):

PR #1960, opened 2022-05-25
[shared] Performance improvement by increasing the check frequency

pras-spydra:
This change reduces the time taken to run full hyperledger network into half. 

With my tests, I used a 3 org setup with 3 ca, 3 orderers, 2 peers, 1 channel & chaincode with EKS. It previously took about 35-40 minutes to complete the run. After the changes, it takes only 15-20 minutes. I believe it will be a big time saver especially for the developers.

**Changes:**
- Flux git.pollInterval changed from `2m` to `1s`
- Delay in tasks in `shared/configuration/roles/check/*` from `30` to `1`
- Changed max value for retry_count variable in `platforms/network-schema.json` to `1200` (1200 secs / 20 minutes)
- Changed samples with `retry_count: 1200` (After the change, every count is one second)

I have also tried to remove the 6 minute hard stop for certificated to become valid in hlf by replacing with a CA pod wait task, but it wasn't successful, so commented out the CA pod wait portion and kept the 6 minute hard stop instead. AFIK, the Fabric CA actually creates the certificates with "valid `from`" time `5 minutes before` the current time, so we could use the certificates right away. I don't understand why in case of bevel it fails. If anyone can shine light on this, would be great.

Signed-off-by: Prasanth Sundaravelu <prasanth.s@spydra.app>


suvajit-sarkar:
Hi @pras-spydra thanks for your PR, In my opinion  the delay and retry count values should be left open for user to tweak as per their system and cluster specs. Reducing the delay between the retries might create congestion in already slow cluster. A better optimization would be to parameterize the delay and have it in network.env section. And then it can be suggested to the users to use delay value as (1200/retry_count) and limit the retry count to min 40 and max 1200 in the network validation

pras-spydra:
@suvajit-sarkar I agree, that makes sense. Will do the changes when I get time.

sownak:
@pras-spydra The git interval for flux is too short. While this is okay when initially setting up the network, it is not okay for a running network, because then unnecessary Kubernetes cpu is spent for fetching details from Git when there has been no changes. If we still want to do this, we can reduce the interval for Flux and then increase it again after the deployment by doing a `helm upgrade`

suvajit-sarkar:
Based on the above comments, this PR is not accepted

PR #1991, opened 2022-06-29
[besu] updated network-proxy-none.yaml

jagpreetsinghsasan:
# Commit to be reviewed

    fix(besu): updated network.yaml 
    
    Primary Changes
    ---------------
    1. Updated the genesis file path in platforms/hyperledger-besu/configuration/samples/network-proxy-none.yaml

 
**Reviewed by**
@suvajit-sarkar


jagpreetsinghsasan:
As we are mounting the bevel repo at `/home/bevel` as mentioned [here](https://hyperledger-bevel.readthedocs.io/en/latest/developer/docker-build.html), if we mention the genesis file present at `/home/user/bevel` path, the ansible playbook fails stating the following path could not be created (using the bevel-build container)

Issue #2039, opened 2022-10-10
CrashLoopBackOff on flux pods while reseting other network on the same cluster


suvajit-sarkar:
**Describe the bug**
The flux pods goes into CrashLoopBackOff when reset playbook is run for any other network on the cluster

**To Reproduce**
1. Setup any two networks
2. Reset one of the networks
3. See CrashLoopBackOff  error on the other network flux pods

**Expected behavior**
The other network deployments should be unaffected by the reset

**Environment (please complete the following information):**
 - Version flux v2

**Additional context**
Seems to an issue with how flux v2 uninstall is done

vishaalreddy:
I'll take this issue

PR #2077, opened 2022-11-23
[shared] fixed flux uninstall in reset-network.yaml

RaviKiranJanjanam:
**Changelog**
Fixed flux uninstall by removing the individual k8s components and then deleting namespace in reset-network.yaml instead of directly removing the namespace

 

**Reviewed by**
@suvajit-sarkar @sownak 

 

**Linked issue**
#2039

### DIFF
--- a/platforms/hyperledger-besu/configuration/reset-network.yaml
+++ b/platforms/hyperledger-besu/configuration/reset-network.yaml
@@ -28,7 +28,73 @@
 
     # Uninstalling Flux is needed so that everything is clean
     # remove this if not needed
-    - name: Uninstall flux
+    ##Delete the Flux deployments
+    - name: Delete flux deployments
+      k8s:
+        state: absent
+        api_version: v1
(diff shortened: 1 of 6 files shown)


jagpreetsinghsasan (changes requested):
LGTM, just one comment to have individual tasks separated by a new line and in-code comments for each tasks (This will also be helpful in future when we change the dir structure and these in-code comments become to go-to docs for roles)

sownak:
Is this deleting only flux deployments and not the CRDs? because the deletion of the CRDs were causing the problem

RaviKiranJanjanam:
> Is this deleting only flux deployments and not the CRDs? because the deletion of the CRDs were causing the problem

Yes , since deletion of CRD's is affecting the flux deployments in other namespaces

sownak (approved):

jagpreetsinghsasan (approved):

Issue #2036, opened 2022-09-26
feat(shared): Option to enable cert-manager for tls certificate creation


suvajit-sarkar:
### Description
As a developer I want to add option for cert-manger to create the tls proxy certificates so that it simplifies the process of obtaining, renewing and using certificates. 

https://cert-manager.io
Explore cert-manager and the issuer to be used for bevel network env proxy tls certificates. SelfSigned issuer is the current suggestion

### Acceptance Criteria
1. Add code to obtain and use cert-manager created certificates

RaviKiranJanjanam:
I will work on this

RaviKiranJanjanam:
We will be using LetsEncrypt instead of self signed Issuer

PR #2079, opened 2022-12-02
[shared] Develop cert manager2

RaviKiranJanjanam:
Added Cert-manager helm chart
added configuration to install cert-manager helm chart, create letsencrypt clusterissuer and create tls certificate for Ambassador 
updated network.yaml/network-schema.json to have new field called fqdn, which would be needed for successful creation of tls  certificate, if fqdn is not provided , self signed certificate will be created as before

 

**Reviewed by**
@suvajit-sarkar @sownak @jagpreetsinghsasan 

 

**Linked issue**
#2036


jagpreetsinghsasan (changes requested):
Please squash merge commits into one and remove the redundant code like these files
image

Issue #2111, opened 2022-12-28
bug(shared): ambassador gets reinstalled even if it was installed earlier


adityajoshi12:
**Describe the bug**
While running the network, if the ambassador is already installed with set of configuration, playbook tries to reinstall it and overrides the earlier configuration
**To Reproduce**
- install the ambassador
- start the network

**Expected behavior**
There should be some checks which should skip the ambassador installation if it is already installed.

**Environment (please complete the following information):**
 - besu

PR #2128, opened 2023-01-22
[ci-skip] Skip ambassador installation when installed

adityajoshi12:
**Changelog**
This PR has changeset for skipping ambassador installation when it's installed. Optionally, now namespace can also be provided when installing the ambassador.

 

**Reviewed by**
@developer_github_id

 

**Linked issue**
#2111


platforms/shared/configuration/roles/setup/ambassador/tasks/main.yaml
```diff
@@ -7,6 +7,23 @@
...
+- name: Set proxy namespace
+  set_fact:
+    proxy_namespace: "{{ network.env.namespace | default('default') }}"
+
+- name: Check if ambassador installed
+  k8s_info:
+    api_version: v1
+    kind: Service
+    name: ambassador
+    kubeconfig: { { kubeconfig_path } }
```
jagpreetsinghsasan, line 19: Please rename this to ``{{ kubeconfig_path }}`` and similarly to the next line var

jagpreetsinghsasan (changes requested):
We shouldn't skip ambassador installation this simply.
There can be scenarios when there are 2 deployments using the same ambassador, but one ambassador is using a port range from 15010-15020 and then another deployment want to use next 10 ports (so they can simply mention 15010-15030 and ambassador will open those 10 more ports).
Either we check the opened ports are same as well to skip ambassador installation or we just include the namespaced proxy installation feature of this PR.

jagpreetsinghsasan:
Also, try to use the [newer format](https://github.com/hyperledger/bevel/blob/develop/.github/pull_request_template.md) for PR message (and similarly the commit message) (most likely you will start receiving the new PR format once the develop branch is merged into main).

adityajoshi12:
> We shouldn't skip ambassador installation this simply. There can be scenarios when there are 2 deployments using the same ambassador, but one ambassador is using a port range from 15010-15020 and then another deployment want to use next 10 ports (so they can simply mention 15010-15030 and ambassador will open those 10 more ports). Either we check the opened ports are same as well to skip ambassador installation or we just include the namespaced proxy installation feature of this PR.

Thanks for the suggestion, made the changes as per the suggestion. We are now checking if the ports are already opened in the existing installation. if they are found to be present then skip installation else the ambassador.

jagpreetsinghsasan (changes requested):
There are still some inconsistencies within the code.
It does certain operations only when `no ambassador_installed`, for example: the task `Format ambassador ports`. This should still format ports, even if ambassador is installed.

Also once fixed, please create an issue to address this update in other platforms also (one issue to mention that we need to update this for fabric, corda, quorum, indy), so that we can have a track of it.

jagpreetsinghsasan:
@adityajoshi12 I saw you added commit(s) after my last review. If you think its ready for review again, you should re-request for review 
image

platforms/shared/configuration/roles/setup/ambassador/tasks/main.yaml
```diff
@@ -99,59 +123,122 @@
...
+    do
+      if [[ ! ${availablePortRange[*]} =~ ${i} ]]; then
+        echo "false" && exit 0
+      fi
+    done
+    echo "true"
+  register: ports_available
+  when:
+    - network.type != 'indy'
+    - ambassador_installed
```
jagpreetsinghsasan, line 155: Please have a space between tasks

jagpreetsinghsasan (approved):
LGTM, thankyou!

platforms/shared/configuration/roles/setup/ambassador/tasks/main.yaml
suvajit-sarkar: @adityajoshi12 kindly have a look

adityajoshi12: @suvajit-sarkar updated...

PR #2152, opened 2023-02-16
[ci-skip] Fix vault-auth when restart

sownak:
**Changelog**
- remove reviewer token in Vault_Kubernetes chart
- Add vault-auth as system:auth-delegator
- Add checks for certificate creation so that job checks are not carried again (jobs vanish when cluster restarts)
- Remove unnecessary fields from network.yaml samples for raft consensus


suvajit-sarkar (approved):
LGTM

PR #2156, opened 2023-02-21
[ci-skip] updated haproxy chart

suvajit-sarkar:
**Changelog**
- Fix haproxy ingress checks
- Update haproxy-ingress chart

 

**Reviewed by**
@jagpreetsinghsasan @sownak


jagpreetsinghsasan (changes requested):
1. Whats the difference between the haproxy chart (which is removed) vs the haproxy-ingress chart (newly added)?
2. Please add the commit message body as well (right now its just a header)

(To add both the header and the body, `git commit -s` will open the editor, where we can write both the header and the body)

suvajit-sarkar:
> 1. Whats the difference between the haproxy chart (which is removed) vs the haproxy-ingress chart (newly added)?
> 2. Please add the commit message body as well (right now its just a header)
> 
> (To add both the header and the body, `git commit -s` will open the editor, where we can write both the header and the body)

1. Both helm charts are for Ingress controller implementation for haproxy Load balancer  (https://github.com/jcmoraisjr/haproxy-ingress), the helm charts are from here (https://github.com/haproxy-ingress/charts)
The difference is in their version.
2. Need to update the PR, will use packaged chart instead of copying the source code. Will update the commit message

jagpreetsinghsasan (approved):

PR #2191, opened 2023-03-24
[besu] updated reset.yaml

### LINKED ISSUE #2157: bug(shared): intermittent flux installation/uninstallation issues
Description:
As a developer, I want to fix the flux resources going in terminating state forever.

Steps to reproduce the behaviour:
1. Reset and redeploy the network on which you are working like besu, fabric, etc.
2. It will throw an error at the time of terminating the namespace in which flux is installed.
3. Check the status of the namespace using the command `kubectl get ns <name of namespace>`
4. You will find your namespace as `Terminating`

Expected behaviour:
Namespace should be deleted and the reset-network should happen successfully.

Screenshot:
Image-FluxTerminating

**Acceptance Criterion :**

1. Reset working as expected for platforms

- [x] HL Besu
- [x] GoQuorum
- [x] R3 Corda OS
- [x] R3 Corda Enterprise
- [x] HL Indy
- [x] HL Fabric

saurabhkumarkardam:
# Commit to be reviewed
----------------------------------

bug(besu): updated reset.yaml

    Added patching and deletion code for the following kubernetes resources
    1. ClusterRole
    2. ClusterRoleBinding
    3. GitRepository
    4. Kustomization
    5. HelmChart
    6. HelmRelease
    
fixes #2157

### DIFF
--- a/platforms/hyperledger-besu/configuration/reset-network.yaml
+++ b/platforms/hyperledger-besu/configuration/reset-network.yaml
@@ -110,6 +110,54 @@
         - "{{ network['organizations'] }}"
       ignore_errors: yes
 
+    # Delete ClusterRoleList related to flux
+    - name: Delete flux ClusterRole
+      k8s:
+        state: absent
+        api_version: v1
+        kind: ClusterRole
+        namespace: flux-{{ network.env.type }}
+        name: "{{ item[0] }}"
+        kubeconfig: "{{ kubernetes.config_file }}"
+      vars:
+        kubernetes: "{{ item[1].k8s }}"
+      with_nested:
+        - ["crd-controller-flux-sau"]
+        - "{{ network['organizations'] }}"
+      ignore_errors: yes
+
+    # Delete ClusterRoleBinding related to flux
+    - name: Delete flux ClusterRoleBindingList
+      k8s:
+        state: absent
+        api_version: v1
+        kind: ClusterRoleBindingList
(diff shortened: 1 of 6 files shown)


platforms/hyperledger-besu/configuration/reset-network.yaml
```diff
@@ -110,6 +110,54 @@
...
+        - "{{ network['organizations'] }}"
+      ignore_errors: yes
+
+    # Patching and Deleting GitRepository resources
+    - include_role:
+        name: "delete/gitrepositories"
+
+    # Patching and Deleting Kustomization resources
+    - include_role:
+        name: "delete/kustomization"
```
adityajoshi12, line 151: Q: is it tested against multiple networks, where we want to delete only one network and other should remain working fine.

jagpreetsinghsasan: True.
@saurabhkumarkardam please have a check for multiple networks and include the `fixes #issue_number` string at the end of both the issue description (I have updated some of it) and the commit message.

jagpreetsinghsasan (changes requested):
Just need the extended testing

jagpreetsinghsasan:
Once the review comment is incorporated, please re-request for review by clicking here
image

suvajit-sarkar (changes requested):

PR #2192, opened 2023-03-24
[besu] fixed the helm ansible module for TLS cert creation using cert manager

ravikiranjanjanam:
changes:
Removed the helm module while creation of TLS cert by cert manager because we are deploying helm chart using shell module
changed the order of organization.issuer to check if it is defined/undefined and then to check the value if it is default/letsencrypt instead of first checking value and then checking if it is defined

Reviewers:
@suvajit-sarkar @sownak @jagpreetsinghsasan 

Fixes:
hotfix


suvajit-sarkar (approved):
LGTM

jagpreetsinghsasan (approved):

PR #2219, opened 2023-04-20
[ci-skip] Fix remove org errors

sownak:
Fixed following errors
- remove org for Fabric was failing because of latest shared changes, it was removing the flux 
- Update to samples
- Renamed shared files correctly
- Updated Job waiting so that it succeeds even when pods were deleted but jobs were complete
- Update helm uninstall


suvajit-sarkar (approved):
LGTM

jagpreetsinghsasan (approved):

PR #2224, opened 2023-04-28
[ci-skip] flux optimisation

adityajoshi12:
flux optimisation
    - setting up strict flux version while installation
    - skip watching other namespaces


sownak (approved):

jagpreetsinghsasan (approved):
Looks good to me

PR #2225, opened 2023-05-01
[besu] add 'none' as a new option for transaction manager

### LINKED ISSUE #2163: Besu - transaction_manager introduce 'none' for not using privacy
**Is your feature request related to a problem? Please describe.**
For Hyperledger Besu, privacy is a addon feature, which is not supported out of the box. I think we should give that flexibility here as well. The parameter `network.config .transaction_manager` is bound to have two values - "tessera" or "orion", which makes the users to provide a value anyway otherwise network schema validation will break. 

**Describe the solution you'd like**
Introduce option of 'none' for transaction_manager in case somebody doesn't want to use privacy member. 

**Describe alternatives you've considered**
Currently to make it work, pass a value for transaction_manager and its related fields like "tm_version","tm_tls","tm_trust" and "tm_nodes"

saurabhkumarkardam:
### **Commit to be reviewed.**
---
**feat(besu): add 'none' as a new option for transaction manager.**
```
Updated the following files to introduce the new option for transaction manager:
	1. sample network-besu.yaml file.
	2. deployment.yaml file of chart node_besu.
	3. node_besu.tpl template.
	
In addition to the above changes, I also made the following updates:
	1. added a label to the job.yaml file of the generate_ambassador_certs chart
	2. corrected playbook names in the add-new-organization.yaml file
```

fixes #2163

### DIFF
--- a/platforms/hyperledger-besu/charts/generate_ambassador_certs/templates/job.yaml
+++ b/platforms/hyperledger-besu/charts/generate_ambassador_certs/templates/job.yaml
@@ -14,6 +14,7 @@ metadata:
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     app.kubernetes.io/instance: {{ .Release.Name }}
+    app: {{ .Release.Name }}
     {{- if $.Values.labels }}
     {{- range $key, $value := $.Values.labels.deployment }}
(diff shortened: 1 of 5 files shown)


jagpreetsinghsasan (approved):
This looks good to me.
But I have a question here, where are we fixing the operational tasks for the same? Like addition of a new member node in the existing network. If we want to address this in another task, then we need to do that on priority, so we want this feature to be fully implemented.

sownak (approved):

PR #2226, opened 2023-05-01
Added application molecule test in CI

### LINKED ISSUE #738: Add Application molecule tests in GitActions
### Description
As a developer I want to run all Example application Molecule test Scenarios so that the Git Actions flow is complete for Applications.
Add a condition so that the Application tests are run only when the check-in has a "[application]" in the commit message

### Acceptance Criteria
1. GitActions test should run for Sample applications when commit message has "[application]"
2. All Application molecule test scenarios should be run for both supplychain and identity. (this can be separated if the tests take more than 5 minutes to run)
3. Tests should run parallelly for supplychain and identity app

octonawish-akcodes:
fixes #738 

Need reviews @sownak @suvajit-sarkar @jagpreetsinghsasan

### DIFF
--- a/.github/workflows/test_molecule.yml
+++ b/.github/workflows/test_molecule.yml
@@ -187,3 +187,16 @@ jobs:
         cd ./platforms/quorum/configuration
         molecule test -s tessera
       if: contains(github.event.pull_request.title, 'quorum')
+#######################################################################################
+  #Application molecule test
+  application-tests:
+    runs-on: ubuntu-latest
+    if: "contains(github.event.head_commit.message, '[application]')"
+    strategy:
+      matrix:
+        app: [supplychain, identity]
+    steps:
+      - name: Run Molecule tests
+        run: |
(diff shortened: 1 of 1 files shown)


jagpreetsinghsasan (changes requested):
Have you tested this one?
Because how will `cd supplychain` gonna work?
We do not have the applications anymore in the bevel repository. They are hosted [here](https://github.com/hyperledger/bevel-samples)

octonawish-akcodes:
How can I test it?

suvajit-sarkar:
Adding app molecule test would require adding the molecule test cases, the work flow won't do anything otherwise

octonawish-akcodes:
Are there any sample test cases I can get idea from, also where should the test cases be placed in the bevel-sample repo or in this repo?

sownak:
@octonawish-akcodes molecule tests are in `platforms/shared/configuration/molecule` which can be referred. Tests for sample will be in the other repo

Issue #2229, opened 2023-05-03
bug(shared): failing to retry the jobs


adityajoshi12:
**Describe the bug**
Jobs are failing due to retry timeout, even though the kubernetes job is successful and completed.

**To Reproduce**
Run the network and you can see the job waiting and timing out 

**Expected behavior**
Once the job is completed, it should proceed to the next ansible task

**Screenshots**
NA

**Environment (please complete the following information):**
- branch used `develop`

**Additional context**
The evaluation conditions need to be checked.

suvajit-sarkar:
The issue should be fixed with this merged PR https://github.com/hyperledger/bevel/pull/2228, can you confirm you have the latest code ?

adityajoshi12:
The issue is with the condition used `component_data.resources[0].status.succeeded` where `succeeded` is not present in the response.

```
            "status": {
                "conditions": [
                    {
                        "lastProbeTime": null,
                        "lastTransitionTime": "2023-05-03T11:28:05Z",
                        "reason": "PodCompleted",
                        "status": "True",
                        "type": "Initialized"
                    },
                    {
                        "lastProbeTime": null,
                        "lastTransitionTime": "2023-05-03T11:28:07Z",
                        "reason": "PodCompleted",
                        "status": "False",
                        "type": "Ready"
                    },
                    {
                        "lastProbeTime": null,
                        "lastTransitionTime": "2023-05-03T11:28:07Z",
                        "reason": "PodCompleted",
                        "status": "False",
                        "type": "ContainersReady"
                    },
                    {
                        "lastProbeTime": null,
                        "lastTransitionTime": "2023-05-03T11:28:05Z",
                        "status": "True",
                        "type": "PodScheduled"
                    }
                ],
                "containerStatuses": [
                    {
                        "containerID": "containerd://05faa5fb2a431bb6bc9824328849efb98d274d54c88d9364dd183bcbe8d85bde",
                        "image": "ghcr.io/hyperledger/alpine-utils:1.0",
                        "imageID": "ghcr.io/hyperledger/alpine-utils@sha256:251b822ecce0da297040b568732aa08e1377a5e44a1abd6892431ddcf5ac5a02",
                        "lastState": {},
                        "name": "vault-kubernetes",
                        "ready": false,
                        "restartCount": 0,
                        "started": false,
                        "state": {
                            "terminated": {
[...]

PR #2230, opened 2023-05-03
[ci-skip]: fix job retries

adityajoshi12:
Change Log

- fix job retries
#2229


sownak:
This change will break existing networks which are deployed with kubernetes 1.23

suvajit-sarkar:
@adityajoshi12 have it as an or condition with existing check if it absolutely necessary on 1.24

Issue #2221, opened 2023-04-24
feat(besu): add a single vault management script for sidecar containers


suvajit-sarkar:
### Description
As a developer I want to abstract all vault curl commands, auth and validation logic to a single vault management script and have helm charts call it when required.

The script can be part of bevel-alpine image or mounted as config map. 

### Acceptance Criteria
1. Helm charts clean of vault curl commands and logic
2. Helm charts to only call vault management script functions

PR #2239, opened 2023-05-12
[besu] Decopule Vault CURD Operations in helm charts 

ravikiranjanjanam:
Changes:

Vault CURD operations have been decoupled and are moved to a script where the vault Script will be created as a configmap in each organizational namespace and will be mounted as a volume in helm charts.

Reviewers:
  @suvajit-sarkar @jagpreetsinghsasan @sownak 

Issues:
#2221


platforms/hyperledger-besu/charts/orion_key_mgmt/templates/job.yaml
```diff
@@ -136,3 +122,7 @@ spec:
                 fi
                 COUNTER=`expr "$COUNTER" + 1`
               fi
+          volumeMounts:
+            - name: scripts-volume
+              mountPath: /scripts/bevel-vault.sh
+              subPath: bevel-vault.sh
```
suvajit-sarkar, line 128: add new line

platforms/shared/charts/vault-script/scripts/bevel-vault.sh
```diff
@@ -0,0 +1,184 @@
...
+                  -H "X-Vault-Token: ${VAULT_TOKEN}" \
+                  -H "Content-Type: application/json" \
+                  -X POST \
+                  -d @${2} \
+                  ${VAULT_ADDR}/v1/${1})
+}
+
+##############################################################################################################################
+################################################## AWS Secret Manager functions starts #######################################
+##############################################################################################################################
```
suvajit-sarkar, line 64: Is this tested ? Remove if not. Use of aws secret manager and akv would need changes in ansible side aswell

suvajit-sarkar (changes requested):

sownak (changes requested):
generally looks good. But, as we now have multiple vaults, I would expect a if condition in all the  Besu charts on which vault to be used.

platforms/shared/charts/vault-script/scripts/bevel-vault.sh
```diff
@@ -0,0 +1,184 @@
...
+                                    --data-raw '{"value": "'"$JSON_PAYLOAD"'"}' )
+}
+
+
+##############################################################################################################################
+################################################## Vault main handler function ###############################################
+##############################################################################################################################
+
+vaultBevelFunc() {
+    if [[ $VAULT_TYPE = "hashicorp" ]]; then
```
suvajit-sarkar, line 123: @sownak the vault conditions are determined here based on the vault type passed into container env variable.

platforms/hyperledger-besu/charts/orion_key_mgmt/templates/job.yaml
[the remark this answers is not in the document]
ravikiranjanjanam: New line added

platforms/shared/charts/vault-script/scripts/bevel-vault.sh
ravikiranjanjanam: Removed the code for aws and Azure secret manager as this is not tested

ravikiranjanjanam:
> generally looks good. But, as we now have multiple vaults, I would expect a if condition in all the Besu charts on which vault to be used.

Removed the code for other vaults like aws secret manager and azure as this is not tested, we are passing Vault_type as environmental variable within the helm chart consuming from helm releases Vaules ,default vault is set to hashicorp and  'if condition' was placed within the shell script. If we are expanding to other vaults, we could have a new field 'type'  in network.yaml under vault section

platforms/hyperledger-besu/charts/generate_ambassador_certs/templates/job.yaml
```diff
@@ -67,57 +71,35 @@ spec:
...
           mkdir -p ${OUTPUT_PATH}
 
-          vault_secret_key=${VAULT_SECRET_ROOTCA}
+          vault_secret_key="${VAULT_SECRET_ENGINE}/${VAULT_SECRET_ROOTCA}"
           echo "Checking certs in ${vault_secret_key}"
-          LOOKUP_SECRET_RESPONSE=$(curl -sS --header "X-Vault-Token: ${VAULT_CLIENT_TOKEN}" ${VAULT_ADDR}/v1/${vault_secret_key} | jq -r 'if .errors then . else . end')
-          CA_PEM=$(echo ${LOOKUP_SECRET_RESPONSE} | jq -r '.data.data["rootca_pem"]' 2>&1)
-          CA_KEY=$(echo ${LOOKUP_SECRET_RESPONSE} | jq -r '.data.data["rootca_key"]' 2>&1)
+          vaultBevelFunc "readJson" ${vault_secret_key} 
+          CA_PEM=$(echo ${VAULT_SECRET} | jq -r '.data["rootca_pem"]' 2>&1)
```
sownak, line 94: which version of vault is this tested on?

sownak (changes requested):

platforms/hyperledger-besu/charts/generate_ambassador_certs/templates/job.yaml
ravikiranjanjanam: Its tested on 1.7.1

suvajit-sarkar (approved):
LGTM, tested ok. New issue will be create to incorporate condition of for vault

platforms/hyperledger-besu/charts/generate_ambassador_certs/templates/job.yaml
suvajit-sarkar: the data.data is required, but its already sending the .data from the script
VAULT_SECRET=$(curl --header "X-Vault-Token: ${VAULT_TOKEN}" ${VAULT_ADDR}/v1/${1} | jq -r 'if .errors then "null" else .data end')

sownak (approved):

PR #2240, opened 2023-05-12
[substrate] Add support for Substrate based DSCP platform

sownak:
This PR makes the following changes
- New platform `substrate` is introduced
- Used `dscp-node` as the substrate node template which will be used in examples as well
- Adds image automation for flux with flux version upgrade
- Fixes formatting errors in docs so Maintainers are listed in readthedocs
- A sample github actions file to deploy a DLT network with Bevel on GCP


jagpreetsinghsasan (approved):
A new platform, lots of small unnoticed bugfixes and new features like .yamlint (I didnt know about this, was much required though)
Looks good to me (I will have to read substrate and try some deployment around it to actually comment on the platform specifics itself, but looks fine overall)

PR #2262, opened 2023-05-26
[substrate] Update chart readme's

sownak:
Added readme and updated Chart.yaml so that the helm charts can be made available as a repo.


suvajit-sarkar (approved):
LGTM

Issue #2273, opened 2023-05-31
feat(fabric): Remove hardcoding of 8443 port and introduce an OrdererAddress


sownak:
### Description
Currently the port 8443 is hardcoded in many places in Fabric. This needs to be removed, and also an OrderAddress like PeerAddress should be introduced which will make it unnecessary to use the proxy check when determining the orderer or peer address.

### Acceptance Criteria
1. 8443 is removed from Fabric code
2. All address should refer to peerAddress and new field OrdererAddress

saurabhkumarkardam:
I will check this.

PR #2301, opened 2023-06-19
[ci-skip] fix hardcoded port 8443 and introduce the ordererAddress field

saurabhkumarkardam:
### **Commit to be reviewed**
---
**feat(fabric): fix hardcoded port 8443 and introduce the ordererAddress field**

```
This commit removes all occurrences of the hardcoded port and introduces a new field called ordererAddress. This new field will be used to refer to the orderer address, similar to how peerAddress is used for the peer address.

Changes Made:
- Removed all occurrences of port 8443 in the Fabric codebase.
- Introduced the new ordererAddress field for referring to the orderer address.

By making this change, it is no longer necessary to rely on a proxy check when determining the orderer or peer address. This improves flexibility and makes the codebase more maintainable.
```

fixes: #2273


sownak (approved):

suvajit-sarkar (approved):
LGTM

Issue #2310, opened 2023-06-30
feat(fabric): Integrate bevel-operator-fabric with bevel to deploy a Fabric network


sownak:
### Description
As a operator I want to have an option to deploy the Fabric network using bevel-operator-fabric so that I can setup the network faster and without using Flux operator.

This story will only cover initial Fabric network setup and does not cover chaincode deployment.

### Acceptance Criteria
1. Option to use bevel-fabric-operator in network.yaml
2. Network setup using Bevel automation and bevel-operator-fabric
3. Complete network deployment and deletion possible by single command.

PR #2313, opened 2023-07-13
[ci-skip] Add operator support

sownak:
This PR adds support for bevel-operator-fabric in traditional Bevel using Ansible. gitops and vault is optional, but proxy = none is not supported because bevel-operator-fabric itself has istio as mandatory.
This PR adds support till channel creation and joining, chaincode management is not supported.

Solves #2310


suvajit-sarkar (approved):
LGTM

PR #2314, opened 2023-07-14
[besu] use default securityContext

adityajoshi12:
```
This commit implements the following changes:
- use the securityContext in the besu and validator nodes.

These changes enable besu and validator nodes to run with default securityContext.
```

### DIFF
--- a/platforms/hyperledger-besu/configuration/roles/create/helm_component/templates/node_besu.tpl
+++ b/platforms/hyperledger-besu/configuration/roles/create/helm_component/templates/node_besu.tpl
@@ -128,3 +128,8 @@ spec:
     metrics:
       enabled: {{ peer.metrics.enabled | default(false)}}
       port: {{ peer.metrics.port | default(9545) }}
+
+    securityContext:
+      runAsUser: 1000
+      runAsGroup: 3000
+      fsGroup: 3000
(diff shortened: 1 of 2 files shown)


suvajit-sarkar:
Do we need to add it in the tpl files ? The Besu node and validator charts have default security context in their values.yaml.
https://github.com/hyperledger/bevel/blob/a9210f7003cf4b6b94bfb353cd4e1bebf152b430/platforms/hyperledger-besu/charts/node_validator/values.yaml#L184

adityajoshi12:
> Do we need to add in tpl files ? The Besu node and validator charts have default security context in their values.yaml.
> 
> https://github.com/hyperledger/bevel/blob/a9210f7003cf4b6b94bfb353cd4e1bebf152b430/platforms/hyperledger-besu/charts/node_validator/values.yaml#L184

Yes, we need to add them, as a new values file for the chart is generated based on the corresponding `.tpl` file.

suvajit-sarkar:
The develop branch charts have the security context added as default in charts values, if this needs to be overridden, change the values.yaml in the chart folder. https://github.com/hyperledger/bevel/blob/d8f2ea376b970541a4d9060c101b5c88eed52f06/platforms/hyperledger-besu/charts/node_besu/values.yaml#L200

Defining it again in tpl is not required. the values used in chart is combined value of values.yaml and the release values

PR #2315, opened 2023-07-17
[ci-skip] simplify Storage Class generation with a common Helm Chart

### LINKED ISSUE #2306: feat(shared): dynamically generate storage class through a single template file based on different cloud provider
As a developer, I want to simplify our project's process of generating storage classes for various cloud providers.

**Issue:**
Currently, we encounter code duplication and maintenance overhead due to the presence of separate template files for generating storage classes across various cloud providers.

---

**Solution:**
However, we now want to dynamically generate a storage class through a single template file based on the various cloud providers.

For instance, if the cloud provider is AWS, the storage class should be generated with an EBS CSI AWS provisioning plugin, and if the cloud provider is GCP, the storage class should be generated with a GCP PD plugin.

To achieve this, we need to create a new Helm chart for the storage class that will have a single template file. We also need to create a Helm Release file template to configure the new Helm chart based on the various providers.

---

**Additional Context:**
This issue will track the development process of dynamically generating the storage class in a single template file/single helm chart and values based on the different cloud providers.

**Fix Required**
Instead of using gitOps to create helm release, use helm install from ansible

---

The following platforms require fixes for this issue:
- [x] Hyperledger-Fabric
- [x] Hyperledger-Besu
- [x] R3-Corda-Ent
- [x] GoQuorum
- [x] R3-Corda
- [x] Hyperledger-Indy
- [ ] Substrate

saurabhkumarkardam:
### **Commit to be reviewed**
---

**feat(shared): simplify Storage Class generation with a Shared Helm Chart**

```
Simplifies the process of generating storage classes by utilizing a shared Helm chart. The four platforms that have been modified to make them compatible with the shared "storage_class" Helm Chart are:
 • Hyperledger-Fabric
 • R3-corda-Ent
 • Hyperledger-Besu
 • Quorum

The following changes have been implemented:
 • Updated variables to align with the shared Helm chart.
 • Introduced common variables to streamline the configuration of storage classes across different platforms.
 • Updated the HR file responsible for configuring the shared "storage_class" Helm chart.

These changes will simplify the generation of various storage classes by utilizing a single template file. Additionally, they will improve compatibility with the shared "storage_class" Helm chart across multiple platforms.
```

fixes #2306


suvajit-sarkar (approved):
LGTM

sownak (approved):

PR #2336, opened 2023-08-02
[ci-skip] Updated docs for minikube

sownak:
- updated docs for minikube
- MacOS tested with HyperKit as VirtualBox has issues
- Removed "GIT_RESET_PATH" as it slows down minikube deployment


suvajit-sarkar (approved):
LGTM

PR #2342, opened 2023-08-11
feat(besu): added support for azure storageclass for Hyperledger Besu

alanna-demuro:


saurabhkumarkardam:
Hi **alanna-demuro**,

Thank you for your feedback. I have reviewed your changes to the StorageClass related code and I agree that your way of doing it is correct. However, we already support this feature where we can use any cloud with any available provisioner and with any parameter of your choice.

The changes that you have made to the StorageClass can also be achieved by editing the values.yaml file at the following path: https://github.com/hyperledger/bevel/blob/develop/platforms/shared/charts/storage_class/values.yaml. Here are the steps on how to make the necessary changes:

```
Step 1: Update the cloud_provider field to azure.
Step 2: Update the provisioner field to disk.csi.azure.com.
Step 3: Add a new list item to the parameters field with the following values:
        azure:
          skuName: StandardSSD_LRS
          kind: Managed
```

I have attached a screenshot of the values.yaml file with the changes that I have mentioned above.
image

You can also check out the Readme.md file for the StorageClass Helm chart at the following path: https://github.com/hyperledger/bevel/blob/develop/platforms/shared/charts/storage_class/README.md. This file provides more detailed information on how to configure the StorageClass.

Note: As we already support this feature, so we won't be able to merge the PR in our code. Thank you for your understanding.

suvajit-sarkar:
@alanna-demuro thanks for your contribution
please add the azure provision details in the file platforms/shared/charts/storage_class/templates/_helpers.tpl
and parameters the azure parameters in the file platforms/shared/charts/storage_class/values.yaml as show in the screenshot above

platforms/hyperledger-besu/configuration/roles/create/k8_component/templates/azure-storageclass
```diff
@@ -0,0 +1,11 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: {{ sc_name }}
+provisioner: disk.csi.azure.com
+parameters:
```
suvajit-sarkar, line 6: add the parameters in platforms/shared/charts/storage_class/values.yaml as shown in the screen shot. this template file is not required

suvajit-sarkar (changes requested):

platforms/hyperledger-besu/configuration/roles/create/k8_component/templates/azure-storageclass
alanna-demuro: @suvajit-sarkar thank you for following up, I had actually tried these steps already before trying the work around, but the creation of the storage class failed.

I just tried again and the Azure storage classes are still failing to create when I use the new charts.

suvajit-sarkar: The issue is with allowedTopologies, make it empty in values.yaml for azure, it should work.

PR #2351, opened 2023-08-23
feat(shared): Simplifying storage class creation with helm module 

sailajakommineni:
This pull request addresses to refactor the current implementation of creating helm releases using GitOps to a more flexible approach by leveraging the ansible helm module and moving the creation of storage class roles to a share folder.

The six platforms that have been modified to make them compatible with the shared "storage class" charts and roles are:
   • Hyperledger-Besu
   • Hyperledger-Fabric
   • R3-corda
   • R3-corda-Ent
   • Quorum
   • Hyperledger- Indy

Changes Made:

Instead of using gitOps to create helm release, used helm module from ansible by creating jinja(j2) template. Moved the creation of storage class roles to a shared folder to use them across all the platforms.

Fixes: #2306

### DIFF
--- a/platforms/shared/configuration/roles/delete/k8s_resources/tasks/main.yaml
+++ b/platforms/shared/configuration/roles/delete/k8s_resources/tasks/main.yaml
@@ -23,3 +23,9 @@
     kubeconfig: "{{ kubernetes.config_file }}"
     namespace: "{{ organization_ns }}"
     name: "{{ organization_ns }}"
+    
+# Delete the existing build directory for StorageClass
+- name: Delete build directory for storageclass
+  file:
+    path: "{{ playbook_dir }}/../../../platforms/shared/configuration/build"
+    state: absent
--- a/platforms/shared/configuration/roles/setup/storageclass/tasks/main.yaml
+++ b/platforms/shared/configuration/roles/setup/storageclass/tasks/main.yaml
@@ -1,0 +1,68 @@
+##############################################################################################
+#  Copyright Accenture. All Rights Reserved.
+#
+#  SPDX-License-Identifier: Apache-2.0
+##############################################################################################
+
+#############################################################################################
+# This role creates value files for storage class
+#############################################################################################
+#############################################################################################
+
+- set_fact:
+    cloud_provider: "{{ org.cloud_provider }}"
+    
+# Check storageclass exists already
+- name: Check if storageclass exists
+  include_role:
+    name: "{{ playbook_dir }}/../../shared/configuration/roles/check/k8_component"  
+  vars:
+    component_type: "StorageClass"
+    component_name: "{{ sc_name }}"
+    kubernetes: "{{ org.k8s }}"
+    type: "no_retry"
+
+- name: Set the storageclass check result to a local variable
+  set_fact:
+    storageclass_state: "{{ result }}"
+
+#############################################################################################
+# create the build directory for storageclass
+- name: Create build directory
+  file:
+    path: "{{ playbook_dir }}/../../../platforms/shared/configuration/build"
+    state: directory
+
+#############################################################################################
+# Creation of the value file for storage class
+- name: Create value file for storage class
+  template:
+    src: "{{ playbook_dir }}/../../../platforms/shared/configuration/roles/create/shared_helm_component/templates/storage_class.tpl"
+    dest: "{{ playbook_dir }}/../../../platforms/shared/configuration/build/{{ sc_name }}-storageclass.yaml"  
+  when: storageclass_state.resources|length == 0
+
+#############################################################################################
+# Create storageclass using helm chart
+- name: Create storageclass using helm chart
+  kubernetes.core.helm:
+    name: "{{ sc_name }}" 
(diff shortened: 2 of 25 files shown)


platforms/shared/configuration/roles/create/shared_helm_component/templates/storage_class.j2
```diff
@@ -0,0 +1,20 @@
+metadata:
```
sownak, line 1: why is this suddenly a .j2 file. Please convert to tpl

platforms/shared/configuration/roles/delete/k8s_resources/tasks/main.yaml
```diff
@@ -18,14 +18,14 @@
 # Remove an existing Storage Class
 - name: Remove an existing Storage Class of {{ organization }}
   vars:
-    provider: "{{ organizationItem.cloud_provider | lower }}"
-    storageclass_name: "{{ organization }}-{{ provider }}-storageclass"
+    cloudProvider: "{{ organizationItem.cloud_provider | lower }}"
+    sc_name: "{{ organization }}-{{ cloudProvider }}-storageclass"
```
sownak, line 22: Storage class should not be removed in delete

platforms/shared/configuration/roles/setup/storageclass/tasks/main.yaml
```diff
@@ -0,0 +1,110 @@
...
+    name: "{{ playbook_dir }}/../../shared/configuration/roles/check/k8_component"  
+  vars:
+    component_type: "StorageClass"
+    component_name: "{{ sc_name }}"
+    kubernetes: "{{ org.k8s }}"
+    type: "no_retry"
+
+- name: Set the storageclass check result to a local variable
+  set_fact:
+    storageclass_state: "{{ result }}"
```
sownak, line 23: Dont think you need two variables for the same "result" use one var

platforms/shared/configuration/roles/setup/storageclass/tasks/main.yaml
```diff
@@ -0,0 +1,110 @@
...
+  template:
+    src: "{{ playbook_dir }}/../../../platforms/shared/configuration/roles/create/shared_helm_component/templates/storage_class.j2"
+    dest: "{{ playbook_dir }}/../../../platforms/shared/configuration/build/{{sc_name}}-storageclass.yaml"
+  vars:
+    component_type: orderer
+    type: "{{ cloudProvider }}-orderer"
+    values_file: 
+      - "{{ playbook_dir }}/../../../platforms/shared/configuration/build/{{ sc_name }}-storageclass.yaml"
+  when: 
+  - component_type == 'orderer' and get_sc.resources|length == 0
```
sownak, line 53: don't think orderer and peer distinction is necessary now that the storageclass name is same for both

platforms/shared/configuration/roles/setup/storageclass/tasks/main.yaml
```diff
@@ -0,0 +1,110 @@
...
+  when: 
+  - component_type == 'peer' and get_sc.resources|length == 0
+  - network.type == 'fabric'
+
+#############################################################################################
+# This task creates the storage class for when using operator 
+- name: Create the storagelass when using operator
+  kubernetes.core.k8s:
+    state: present
+    src: "{{ playbook_dir }}/../../../platforms/shared/configuration/roles/create/shared_helm_component/templates/storage_class.j2"
```
sownak, line 76: Has this been tested? Does k8s module work with template directly?

platforms/shared/configuration/roles/setup/storageclass/tasks/main.yaml
```diff
@@ -0,0 +1,110 @@
...
+  tags:
+    - notest
+
+#############################################################################################
+ # Install the helm chart to create storageclass
+- name: Install Helm Chart for Create storageclass
+  community.kubernetes.helm:
+    name: storageclass
+    chart_ref: "{{ playbook_dir }}/../../../platforms/shared/charts/storage_class"
+    release_namespace: "{{ component_ns }}"
```
sownak, line 92: use default namespace as we dont want the storageclass to be deleted when the namespace is deleted

platforms/shared/configuration/roles/create/shared_helm_component/templates/storage_class.j2
suvajit-sarkar: I think we should move back to .tpl for templates, though from template engine perspective it doesn’t care about extension. Ansible seems go with .j2 but .tpl makes more sense to me as it says something about its use

suvajit-sarkar (approved):
LGTM

platforms/shared/configuration/roles/create/shared_helm_component/vars/main.yaml
```diff
@@ -6,3 +6,5 @@
 
 helm_templates:
   vault_k8s_mgmt: vault_kubernetes_job.tpl
+  awsstorageclass: storage_class.tpl
```
suvajit-sarkar, line 9: why are these changes required ?

sownak (approved):

PR #2356, opened 2023-08-25
[shared] update ambassador to latest ambassador edge-stack

suvajit-sarkar:
**Note**
Changes currently test and enabled for **Besu platform** only

**Change log**
- Update ambassador to latest **ambassador edge-stack (aes)** version 3.7.2
- Orion TM support removed, as Orion support is available via Tessera TM
- RPC url will no longer support custom ports and will use http port 80 only, this will save some ports 
- RLPX protocol does not work over http port, and by default does not support SNI, will continue to use custom ports
- Removed creation of self signed tls cert for ambassador tls termination, instead use aes create default one
- network.env.ambassadorPorts.ports should be an array or list for besu. AJV needs to be ignored for this till aes is enabled on all platforms. Suggest using portRange meanwhile 
- network.env.namespace changed to network.env.proxy_namespace

**TODO**
- Enable aes for other platforms
- Test aes with cert manager


sownak (approved):
LGTM

PR #2374, opened 2023-09-22
[fabric] remove ambassador references

sownak:
Remove all references to Ambassador for Fabric as we will never use Ambassador for Fabric because Ambassador does not support ssl-passthrough


suvajit-sarkar (approved):
LGTM

Issue #2375, opened 2023-09-22
Network Schema doesnt support fabric node chaincode


cw6365:
**Describe the bug**

The network schema doesn't have node in the enum so fails the build, it says on the docs that it's supported? Is it?

https://github.com/hyperledger/bevel/blob/28ff161b268d5869c131be5c81c86eaebe602e3e/platforms/network-schema.json#L677

I'll add a PR if it is.

sownak:
Seems like a miss, please send a PR. I will assign this to you.

PR #2376, opened 2023-09-22
[fabric] Added 'node' option to fabric_chaincode network schema

cw6365:
The network schema doesn't have node in the validating enum so fails the build - #2375


sownak:
@cw6365 Thanks for the PR. Few updates though
- Raise PR to develop branch
- Sign the commit

PR #2377, opened 2023-09-23
[fabric] Added 'node' option to fabric_chaincode network schema

cw6365:
The network schema doesn't have node in the validating enum so fails the build - #2375


cw6365:
@sownak i had to recreate to add my signature, is that all correct?

jagpreetsinghsasan (approved):
LGTM

sownak:
> @sownak thanks, i had to recreate to add my signature, is that all correct now?

No, the DCO is still failing. Please check the actions in the failure message

cw6365:
added DCO, thanks

sownak (approved):

PR #2382, opened 2023-10-06
bug(r3cordaos) vault k8s release files not getting removed on reset

suvajit-sarkar:
**Fix**
- Doorman, NMS and notary vault k8s release files will be created under the org name folder, which is deleted on reset.
- Added missing variables in `setup vault kubernetes` task, in add-new-channel playbook.
- Removed service account and clusterolebinding checks. This fixes nms and notary vault k8 jobs getting skipped.

### DIFF
--- a/platforms/shared/configuration/roles/setup/vault_kubernetes/tasks/main.yaml
+++ b/platforms/shared/configuration/roles/setup/vault_kubernetes/tasks/main.yaml
@@ -114,9 +114,7 @@
     policydata: "{{ lookup('file', '{{ playbook_dir }}/build/vault-crypto-{{ component_type }}-{{ name }}-ro.hcl') }}"
     create_serviceAccount: "{{ check_serviceAccount }}"
     create_clusterRoleBinding: "{{ check_clusterRoleBinding }}"
-    values_dir: "{{ playbook_dir }}/../../../{{gitops.release_dir}}/{{ name }}"
-  when:
-  - check_serviceAccount or check_clusterRoleBinding
+    values_dir: "{{ playbook_dir }}/../../../{{ gitops.release_dir }}/{{ org_name | default(name) }}"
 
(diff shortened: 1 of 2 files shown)


sownak (approved):

platforms/shared/configuration/roles/setup/vault_kubernetes/tasks/main.yaml
```diff
@@ -61,44 +61,6 @@
     KUBECONFIG={{ kubernetes.config_file }} kubectl config view --minify | grep server | cut -f 2- -d ":" | tr -d " "
   register: kubernetes_server_url
 
-# Check if ClusterRoleBinding created
-- name: "Checking for the {{ component_ns }}-role-tokenreview-binding"
-  include_role:
-    name: "{{ playbook_dir }}/../../shared/configuration/roles/check/k8_component"
-  tags:
-    - notest
```
sownak, line 69: Are these deletions necessary?

suvajit-sarkar: Had to remove the checks, as the problem is with corda platform components (doorman, nms and notary) has there own vault auth policy as they have different secret engine path but they belong to same org. So if we keep the check doorman vault k8s job will create service account and clusterrole binding and then nms and notary jobs will be skipped as it finds the sa and clusterrole already present.

suvajit-sarkar: Alternative approaches would be to have one vault auth policy by having these components under one secret engine path and run vault k8s job one time for the org, or create different service account and clusterrolebinding for doorman, nms and notary

suvajit-sarkar: Update: Have added the checks back, as it was used in the task variables, have removed the when condition.

PR #2394, opened 2023-10-19
feat(shared): add missing files for OS-specific command handling

saurabhkumarkardam:
### **Quick fix commit to be reviewed**
---

**feat(shared): add missing files for OS-specific command handling**

```
Description:
- In the PR titled "feat(shared): enable OS-specific command handling in Helm charts," I forgot to include certain files.
- I thought that changing the name of the Helm chart would override the updated file only, not the entire chart, but I was wrong.
- This PR will address the missing file and fulfill the purpose of the original PR.
```

fixes #quick-fix


suvajit-sarkar (approved):
LGTM

PR #2395, opened 2023-10-20
[fabric] Update documentation

### LINKED ISSUE #2391: bug(fabric) Update documentation
**Describe the bug**
As a developer I want to update the values of the read.me and values.yaml files of the charts to more realistic values than those currently used on the network.
Also correct some errors in the documentation such as:
https://github.com/hyperledger/bevel/blob/develop/platforms/hyperledger-fabric/charts/create_channel/README.md?plain=1#L24
This line should be the following:
https://github.com/hyperledger/bevel/blob/develop/platforms/hyperledger-fabric/charts/create_channel

mgCepeda:
Primary Changes
--------------
 1. Updated README.md and values.yaml files with more realistic values than those currently used on the network
 2. Added some values that weren't described

fixes #2391


suvajit-sarkar (approved):
LGTM

PR #2403, opened 2023-10-27
[fabric] Custom container registry support

### LINKED ISSUE #2386: Custom container registry support
**Describe the bug**
Currently the Bevel framework has an option to specify the container registry username and password. This information however does not help unless somebody manually updates the default value spec of every container image.

Container images are hardcoded in helm charts as default values. For example: https://github.com/hyperledger/bevel/blob/28ff161b268d5869c131be5c81c86eaebe602e3e/platforms/hyperledger-fabric/charts/peernode/values.yaml#L17
The update on this default value is possible through the HelmRelease, however those updates are not consistent. For example: https://github.com/hyperledger/bevel/blob/28ff161b268d5869c131be5c81c86eaebe602e3e/platforms/hyperledger-fabric/configuration/roles/create/peers/tasks/nested_main.yaml#L27 does not use the specified container image registry. The image for couchdb in the same file does.

**To Reproduce**
Regular workflow of Bevel with a custom container registry.

**Expected behavior**
The ideal solution is to use the container registry URL consistently for all images. Then each release to have a list of pre-requisite images that somebody needs to have to use the Bevel project. Use the container image repository information to arrive at the value spec for the container images.

**Screenshots**
NA

**Environment (please complete the following information):**
 - applicable to all environments.

**Additional context**
None

mgCepeda:
Primary Changes
--------------
 1. All tpl have been modified so that all images are under the same registry
 2. All image references in the roles have been removed
 3. Fixed an error that occurs in Fabric deployment 1.4.8
 4. The values.yaml and readme files have been updated with the new values of the images

fixes #2386


suvajit-sarkar (approved):
LGTM

PR #2427, opened 2023-11-09
[r3-corda-ent)] fix chart name and linting issue

saurabhkumarkardam:
### **Commit to be reviewed**
---

**bug(r3-corda-ent): fix chart name and linting issue**

```
Fixed the chart name at following locations:
- Ansible role.
- Template file (.tpl) file.
- Helm Chart name itself.

Linting issue:
- Fixed for the "corda-ent-node" Helm chart.
```

fixes: quick-fixes


platforms/r3-corda-ent/charts/corda-ent-node/values.yaml
```diff
@@ -173,7 +173,7 @@ networkServices:
 # external firewall (Bridge and Float)
 firewall:
   # enabled is a boolean value for the above function
-  enabled:
+  enabled: false
```
saurabhkumarkardam, line 176: Helm linting: cannot set a boolean variable to empty for independent Helm chart deployment.

suvajit-sarkar (approved):

PR #2441, opened 2023-12-04
Rename the Shared charts as per Bevel conventions

sownak:
- Rename storageclass chart
- Rename vault-mgmt chart
- Rename imageautomation chart


suvajit-sarkar (approved):
LGTM

PR #2471, opened 2024-01-09
[besu] Updated Charts with dependencies and hooks

sownak:
Release 1.0.0 Changes for Besu
- Nodekey and other secret generation moved from separate chart to Hooks.
- Tessera secret generation also moved from separate chart to Hooks.
- TLS cert Generation for use with proxy is now a dependent chart called `besu-tlscert-gen`.
- `bevel-scripts`, `bevel-storageclass` and `bevel-vault-mgmt` are dependent Charts for new Chart `besu-genesis`.
- `besu-tessera-node` is dependent chart for `besu-node` depending on privacy flags.
- Independent jobs will no longer use Flux/Gitops
- All deployments will start with a `genesis` chart which does the startup
- The first organization in the network.yaml must be a validator

Deprecations:
- `ethash` and `raft` consensus
- Add member/validators are not tested with current format. Will be released later.
- Generate crypto playbook also removed until the need for this playbook is established.
- `letsencrypt` certificates and `Prometheus` needs to be tested again.


.github/workflows/release_charts.yml
```diff
@@ -38,6 +38,7 @@ jobs:
       - name: Install additional Helm repos
         run: |
           helm repo add bitnami https://charts.bitnami.com/bitnami
+          helm repo add bevel https://sownak.github.io/bevel
```
suvajit-sarkar, line 41: Do we require this helm repo ?

.gitignore
```diff
@@ -38,3 +38,6 @@
 **/.vscode/
 # custom config templates
 *_custom.tpl
+**/charts/*.tgz
+**/files/*.json
+requirements.lock
```
suvajit-sarkar, line 43: Kindly add new line

platforms/substrate/configuration/deploy-network.yaml
```diff
@@ -55,7 +55,6 @@
       name: "{{ playbook_dir }}/../../shared/configuration/roles/setup/scripts"
     vars:
       namespace: "{{ org.name | lower }}-subs"
-      network_type: "{{ network.type | lower }}"
```
suvajit-sarkar, line 58: Saw this change in other files as well. Did we change anything related to network.type or was this not used ?

suvajit-sarkar (changes requested):
LGTM. Just few files need new line addition and network schema needs to be updated.

platforms/substrate/configuration/deploy-network.yaml
sownak: This is not used

.gitignore
sownak: Yes added for others.

.github/workflows/release_charts.yml
sownak: Yes for now, will have to be updated to Hyperledger once all charts are done

suvajit-sarkar (approved):
LGTM

PR #2472, opened 2024-01-10
[indy] Fix for indy-key-mgmt container to build correctly

sownak:
Emergency fix for indy-key-mgmt build error


suvajit-sarkar (approved):
LGTM

Issue #2473, opened 2024-01-11
policy_type variable undefined


Linkadi98:
**Describe the bug**
When I run `run.sh` script, task Create policy for Access Control shows error `The error was: 'policy_type' is undefined`

**To Reproduce**
1. Prepare a cluster by Minikube with docker driver
2. Prepare Vault by Helm 
3. Follow instruction to setup build folder
4. Build docker image and run container

**Expected behavior**
Go to the next step

**Screenshots**
<img width="1526" alt="image" src="https://github.com/hyperledger/bevel/assets/36079989/60b411c5-3ae9-4de3-806f-562d3e9d8f81">

**Environment (please complete the following information):**
 - OS: Ubuntu server with public ip
 - Version 22.04 LTS 

**Additional context**
I follow the instruction in the `tutorial` folder to setup environment

Linkadi98:
As the log says, the `policy_templates` missing `policy_type` as below

`policy_templates:
        quorum: quorum-policy.tpl
        besu: besu-policy.tpl
        cordaos: cordaos-policy.tpl
        fabric: fabric-policy.tpl
        r3-corda-ent: r3-corda-ent-policy.tpl
`

This config doesn't contain `policy_type` field, I also cannot find the fabric-policy.tmp file in anywhere

suvajit-sarkar:
the policy_type variable is defined in the calling role. In your case, I assume you trying to deploy Fabric network. The policy_type is defined here
https://github.com/hyperledger/bevel/blob/487f659b7e3637db3f98383b50ebe16b158392c8/platforms/hyperledger-fabric/configuration/deploy-network.yaml#L62

suvajit-sarkar:
What branch/tag are you using ?

Linkadi98:
I saw policy_type in this file before, but the log says it is not defined, so how do I fix this error? Before the latest release, I have never seen this error 😗

Linkadi98:
> What branch/tag are you using ?

I use develop branch with the latest code

Linkadi98:
I just check the code in main branch and everthing is okay, the issue only appears at develop branch

suvajit-sarkar:
Looks like valid issue on dev branch

Linkadi98:
Today I try to run network with minikube driver=none, which means I use directly the host to make a cluster and the issue apprears after this config, and I also use the latest code at main branch.
@suvajit-sarkar

saikumarbommakanti:
I am working on this

PR #2479, opened 2024-01-12
[besu] Update for add new member node

sownak:
- Merge main to develop
- Tested and updated for adding new member node for Besu and Quorum
- Added fix for #2473


suvajit-sarkar (approved):
LGTM

PR #2501, opened 2024-01-30
Moving Aries Agent charts from Hyperledger Labs to Bevel

monar24:
Our team at superlogic proposed a hyperledger lab for our aries-agent charts. We were asked to raise a PR for contributing this part to Bevel. Previous conversation may be viewed here: https://github.com/hyperledger-labs/hyperledger-labs.github.io/pull/259


suvajit-sarkar (approved):
LGTM

platforms/hyperledger-aries/chart/values.yaml
```diff
@@ -0,0 +1,164 @@
...
+  group: 'vc-authn-oidc'
+  name: 'vc-authn-oidc'
+  role: 'agent'
+
+controller:
+  role: 'controller'
+  webhookBaseURL: ''
+
+deploy:
+  image:
```
sownak, line 21: can we use `image.repository` and `image.tag` format for the images

platforms/hyperledger-aries/chart/values.yaml
```diff
@@ -0,0 +1,164 @@
...
+    name: 'ghcr.io/hyperledger/aries-cloudagent-python'
+    ## Modify aca-py agent version: 
+    tag: 'py3.9-0.9.0'
+  name:
+    prefix: 'aries-agent'
+    suffix: ''
+
+ingress:
```
sownak, line 29: Bevel calls ingress and proxy, so can this be moved to key called `global.proxy`

sownak (changes requested):
How is multi-cloud supported?

suvajit-sarkar:
Hi @monar24 i think it will make more sense to put the Aries helm charts in https://github.com/hyperledger/bevel-samples under identity rather having under Bevel DLT platforms. Or maybe we put under Indy platform as an identity solution. ?

monar24:
> https://github.com/hyperledger/bevel-samples



> Hi @monar24 i think it will make more sense to put the Aries helm charts in https://github.com/hyperledger/bevel-samples under identity rather having under Bevel DLT platforms. Or maybe we put under Indy platform as an identity solution. ?

Thanks everyone for the feedback and time reviewing. 

I actually agree with moving it to bevel-samples. Aries agents are deployed in order to interact with an underlying ledger, the ledger being indy by default mostly. Since bevel itself is for deploying the actual networks, I agree that it may be better to move these agent charts there. 

I have raised that PR here: https://github.com/hyperledger/bevel-samples/pull/78

suvajit-sarkar:
Closing this PR as we have moved to https://github.com/hyperledger/bevel-samples/pull/78

swcurran:
Flagging @WadeBarnes about this contribution.

PR #2521, opened 2024-03-04
[corda] Ansible automation with new helm charts

sownak:
- All automation for deploying the supplychain example with new helm charts
- Add Cordapps is pending


suvajit-sarkar (approved):
LGTM

PR #2562, opened 2024-05-17
update(besu,quorum): adapt besu & quorum ansible tasks and template files to charts

alvaropicazo:
This PR contains:

 - Changes in ansible tasks from quorum & besu
 - Changes in tpl files from quorum & ansible
 - Fix in besu configmap for transaction pool max capacity size variable


platforms/hyperledger-besu/charts/besu-node/templates/besu-config-configmap.yaml
```diff
@@ -27,7 +27,7 @@ data:
     node-private-key-file={{.Values.node.besu.privateKeyPath | quote }}
 
     # Transaction Pool
-    tx-pool-layer-max-capacity={{ .Values.node.besu.txPool.maxCapacity }}
+    tx-pool-max-size={{ .Values.node.besu.txPool.maxCapacity }}
```
suvajit-sarkar, line 30: why are we changing this ?

platforms/quorum/charts/quorum-genesis/Chart.yaml
```diff
@@ -1,5 +1,5 @@
-# apiVersion: v2
```
suvajit-sarkar, line 1: why change the apiVersion ?

suvajit-sarkar (commented):
Rest LGTM

platforms/hyperledger-besu/charts/besu-node/templates/besu-config-configmap.yaml
alvaropicazo: Due to the changes of the quorum version. They have renamed that variable

alvaropicazo: Due to the changes of the quorum version. They have renamed that variable

platforms/quorum/charts/quorum-genesis/Chart.lock
```diff
@@ -0,0 +1,9 @@
+dependencies:
+- name: bevel-vault-mgmt
```
sownak, line 2: don't checkin lock files

sownak (changes requested):
Few changes needed

PR #2573, opened 2024-06-03
[fabric] Fix Chart versions and add-cli via helm chart

sownak:
- Chart version update in dependencies
- `add-cli` readme updated with helm install option


suvajit-sarkar (approved):
LGTM

PR #2583, opened 2024-06-18
feat(fabric): update manage-user-certificate playbook to refresh user certificates

alvaropicazo:
This PR contains updates in the manage-user-certificate playbook, as well as minor changes in the ca-tools role to consider the refresh-user-certs variable so whenever this playbook is run, only the user certificates will be affected (updated).


sownak (approved):

suvajit-sarkar (approved):
LGTM

PR #2590, opened 2024-06-28
fabric: hot fix

mgCepeda:
Fixed the execution of the playbook chaincode-ops.yaml when running the playbook add-organization.yaml


suvajit-sarkar (approved):
LGTM

PR #2632, opened 2024-11-13
update maintainers and code reviewers

suvajit-sarkar:


CODEOWNERS
```diff
@@ -1,2 +1,2 @@
-*       @sownak @suvajit-sarkar @adityajoshi12
+*       @sownak @suvajit-sarkar @adityajoshi12 @alvaropicazo
```
sownak, line 1: Alvaro has to be added to the groups first.

suvajit-sarkar: @sownak what group? I invited @alvaropicazo to https://github.com/orgs/hyperledger-bevel/teams/bevel-committers. @alvaropicazo kindly accept the invite. Is that the group ?

suvajit-sarkar: He is getting auto removed. I think there was a page where we have to add his id and commit it, I'm not able find the link. @ryjones would be able to help us with this?

ryjones (approved):

PR #2643, opened 2025-01-15
bug(besu): update sql statements for mysql 8.0+ version

alvaropicazo:
Small changes on SQL Queries due to new version deprecations


alvaropicazo:
@sownak @suvajit-sarkar 
By running the committed lines, a new root user is created for '%' host , as shown in the following screenshot.
Captura de pantalla 2025-01-15 151030
But I wonder if we wanted this or update the root for localhost. When trying to run the following query in the sql script <ALTER USER 'root'@'localhost' IDENTIFIED BY '$pass'> , this error happens.
My gut tells me the one for localhost is the one we really need to update, but not sure if it is needed for privacy transactions executions.
Captura de pantalla 2025-01-15 151250
demouser is correct for '%' host.

sownak:
@alvaropicazo Add a comment about this and then we are good to go

alvaropicazo:
done!

suvajit-sarkar (approved):
LGTM

PR #2645, opened 2025-01-17
bug(besu): update tx pool max capacity argument and values files

alvaropicazo:
- Updated tx-pool-layer-max-capacity argument on besu-configmap
- Updated values files in vault_and_ambassador besu version from 22.10.2 to 23.10.2


suvajit-sarkar (approved):
LGTM

PR #2656, opened 2025-03-31
bug(besu): the orderer chart requires bevel-vault-script config map on minikube

johannww:
Related to issue: https://github.com/hyperledger-bevel/bevel/issues/2653

The problem: the orderer pod always tries to mount bevel-vault-script configmap, but such resource is not created for a minikube deploy.

In this PR, the configmap is always created to be mounted.


alvaropicazo:
Hi @johannww would you mind to sync your fork with the base dev branch and it's latest changes and push again?

sownak (approved):

PR #2675, opened 2025-07-28
(fix): Errors related to duplicate use of ansible var 'register'

sownak:
- also removed aws-auth as dependency because new aws cli tool includes that
- Updated anchorpeer chart to accept - dashed orgname


alvaropicazo (approved):

sownak:
DRAFT- DO NOT MERGE

